PR #4, opened 2021-09-29
Expand CI To Include A Matrix of Rust Versions/LLVM Versions

dave-tucker:
This uses `which` to ensure that we find a path to a "FileCheck|-12" binary, as well as adding LLVM12, LLVM13 and Rustc 1.54, 1.55, beta (1.56) and nightly (1.57) to the matrix.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -5,30 +5,87 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+  schedule:
+    - cron: '00 4 * * *'
 
 env:
   CARGO_TERM_COLOR: always
 
 jobs:
   build:
-
     runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        rust:
+         - "1.54"
+         - "1.55"
+         - "beta"
+         - "nightly"
+        llvm: ["12", "13", "rustc"]
+    name: "rustc: ${{ matrix.rust }}, llvm: ${{ matrix.llvm }}"
 
+    env:
+      CARGO_ARGS: ${{ matrix.llvm == 'rustc' && '--no-default-features --features rust-llvm' || '' }}
+      RUST_BACKTRACE: full
+    
     steps:
-    - uses: actions/checkout@v2
-    - uses: Swatinem/rust-cache@v1
-    - name: Install LLVM
-      run: |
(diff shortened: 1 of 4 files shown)


dave-tucker:
@alessandrod the matrix is complete, the only issue is with enabling the tests without `TESTS_HOST_TARGET` set on nightly. These currently fail because we need `core` to compile the examples, and we don't have a compiled `core` for `bpfel-unknown-none`. As this uses `rustc`, not `cargo`, we can't use `-Zbuild-std=core`, 

I've asked on the Rust discord if anyone has ideas :thinking:

alessandrod:
> @alessandrod the matrix is complete, the only issue is with enabling the tests without `TESTS_HOST_TARGET` set on nightly. These currently fail because we need `core` to compile the examples, and we don't have a compiled `core` for `bpfel-unknown-none`. As this uses `rustc`, not `cargo`, we can't use `-Zbuild-std=core`,
> 
> I've asked on the Rust discord if anyone has ideas 🤔

We could try to make this use `cargo rustc`? https://github.com/alessandrod/bpf-linker/blob/ca1c30f2e248e2570440a02f970dc28b3b97cdac/third-party/compiletest-rs/src/runtest.rs#L1373

dave-tucker:
@alessandrod here's where I got to:

1. I tried [trybuild](https://github.com/dtolnay/trybuild/) which uses cargo, but does support using additional crates (e.g aux-build) and doesn't (currently) support args for target
2. Hacking `compiletest_rs` to use `cargo rustc`, which doesn't work either as cargo rustc doesn't want to compile a single file

Won't make much more progress on this today :cry:

dave-tucker:
ok @alessandrod, i've got the bpf-target tests running for `nightly`, but it requires building from source, so each run takes around 50 minutes :scream:

dave-tucker:
(pushed another update to simplify the workflow and add "rustc" to the llvm matrix, so we test the llvm bundled with rustc)

dave-tucker:
assuming the nightly matrix branches pass, i'm not going to change this workflow again :crossed_fingers: it's as simple as i can make it, and covers everything I think we need.

alessandrod:
> assuming the nightly matrix branches pass, i'm not going to change this workflow again 🤞 it's as simple as i can make it, and covers everything I think we need.

This is amazing thank you! 🙏

PR #8, opened 2021-12-07
Change the type of LLVM Message pointer to libc :: c_char

ken109:
When I tried to install it on Ubuntu, I got the following error.

```
error[E0308]: mismatched types
  --> src/llvm/message.rs:34:36
   |
34 |                 LLVMDisposeMessage(self.ptr);
   |                                    ^^^^^^^^ expected `u8`, found `i8`
   |
   = note: expected raw pointer `*mut u8`
              found raw pointer `*mut i8`

error[E0308]: mismatched types
  --> src/llvm/message.rs:43:52
   |
43 |             let contents = unsafe { CStr::from_ptr(self.ptr).to_str().unwrap() };
   |                                                    ^^^^^^^^ expected `u8`, found `i8`
   |
   = note: expected raw pointer `*const u8`
              found raw pointer `*mut i8`

// Many others
```

The reason was that the arguments of LLVMDisposeMessage and LLVMCreateBinary are libc :: c_char type instead of i8 type.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,7 @@
 #![deny(clippy::all)]
 
+extern crate libc;
+
 mod llvm;
 mod linker;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,7 +1,5 @@
 #![deny(clippy::all)]
 
-extern crate libc;
-
 mod llvm;
 mod linker;
 
--- a/src/llvm/message.rs
+++ b/src/llvm/message.rs
@@ -2,13 +2,14 @@ use std::ffi::CStr;
 use std::fmt;
 use std::ptr;
 
+use libc::c_char;
 use llvm_sys::core::LLVMDisposeMessage;
 
 /// Convinient LLVM Message pointer wrapper.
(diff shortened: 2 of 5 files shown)


src/lib.rs
```diff
@@ -1,6 +1,8 @@
 #![deny(clippy::all)]
 
+extern crate libc;
```
alessandrod, line 3: I don't think this is needed?

src/llvm/message.rs
```diff
@@ -8,7 +8,7 @@ use llvm_sys::core::LLVMDisposeMessage;
 /// Does not own the ptr, so we have to call `LLVMDisposeMessage` to free message memory.
 #[repr(C)]
 pub struct Message {
-    pub ptr: *mut i8,
+    pub ptr: *mut ::libc::c_char,
```
alessandrod, line 11: add a `use libc::c_char;` declaration at the top and use `c_char` here and in other places please

alessandrod (changes requested):
Thank you and sorry for the wait! This slipped under my radar. See my comments.

ken109:
@alessandrod 
Fixed!

alessandrod:
Thank you!

PR #11, opened 2022-01-18
update the git org to aya-rs and CI rust versions

dmitris:
PR update sthe git org to `aya-rs` and adds current CI rust versions.


alessandrod (approved):
Thank you!

PR #12, opened 2022-01-23
update the description of the llvm requirements

dmitris:


dave-tucker (changes requested):
What issue is this hoping to solve?
Per the `llvm-sys` docs:

```
The build scripts will not enforce this compatibility matrix strictly, permitting compilation against any version of LLVM that is at least as new as the crate target version. This is safe in most cases because the LLVM C API is meant to maintain binary compatibility across releases with the exception of when functions are deprecated and later removed. An incompatible LLVM version will generally fail to compile with a link-time error, rather than cause runtime errors. Where versions are known to break binary compatibility, the build script will prevent compilation.
```

So it's likely the correct approach is to stay using `llvm-sys=120" since LLVM13 and 14 can still be used so long as none of the functions we depend on are deprecated.

It's possible to support multiple LLVM versions here by adding a dependency on multiple versions of llvm-sys, but we'd need to use features to ensure they aren't used concurrently.

alessandrod:
Closing this as we depend on LLVM 14 now

PR #13, opened 2022-01-24
Don't strip debug info by default

dave-tucker:
This allows for BTF to be generated!
The original behaviour can be brought back using the wasm-ld compatible
--strip-debug flag

TODO:
- [x] Struct Names are "illegal" - `PerfEventArray<syscall_digest_common::SyscallLog>`

Fixes:  #1 

Signed-off-by: Dave Tucker <dave@dtucker.co.uk>


tamird:
Let's close this in favor of #35?

PR #18, opened 2022-07-13
Add a cli flag to keep btf

vlad9486:
Needed to write bpf code in Rust without C. Using this flag with `--disable-memory-builtins` doesn't crash.


alessandrod:
Can you provide a repro example that crashes without `--disable-memory-builtins` and doesn't when you pass it?

alessandrod:
Also, see https://github.com/aya-rs/bpf-linker/pull/13/files

vlad9486:
Sorry, the example is not minimal.

Remove this line
https://github.com/vlad9486/ebpf-tools/blob/master/example-kprobe/build.rs#L26

```
cargo install bpf-linker --git https://github.com/vlad9486/bpf-linker --branch keep-btf
cargo +nightly build -p example-kprobe
sudo ./target/debug/example-user
```

It will be very good if I can build bpf application with memory builtins.

PR #34, opened 2023-04-14
fix HOSTS_TARGET compilation on mac os

davibe:
HOSTS_TARGET is broken on on macos where the native object code format is Mach-O. This fixes it by ensuring that mach-o files (eg the infamous symbols.o added by rustc during compilation) are skipped

### DIFF
--- a/src/linker.rs
+++ b/src/linker.rs
@@ -515,7 +515,7 @@ fn detect_input_type(data: &[u8]) -> Option<InputType> {
 
     use InputType::*;
     match &data[..4] {
-        b"\x42\x43\xC0\xDE" => Some(Bitcode),
+        b"\x42\x43\xC0\xDE" | b"\xDE\xC0\x17\x0b" => Some(Bitcode),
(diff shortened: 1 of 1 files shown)


alessandrod:
For future reference, my summary of the issue after trying to understand why this is needed:

In https://github.com/aya-rs/bpf-linker/pull/19 I added support for skpping the rustc-generated `symbols.o`, which doesn't contain bitcode. So I added a check to skip valid (elf or archive) inputs that don't contain bitcode.

On macos with `HOSTS_TARGET=1` `symbols.o` is a mach-o file. Without the changes in PR it fails linking early in `link_modules()`  since `detect_input_type()` fails as it doesn't support mach-o. 

So this PR adds mach-o support to `detect_input_type()`, but that's not enough! Object files in macos _do_ have bitcode by default, but native bitcode (aarch64 and x86) not ebpf bitcode. So on macos we want to outright skip mach-o files.

alessandrod (approved):

PR #50, opened 2023-06-01
github: cache rust build products

tamird:
Building rust from source takes our CI time from 3 minutes to 29
minutes. Cache the build products and pray that this reduces the time.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -49,6 +49,13 @@ jobs:
           path: rust
           submodules: recursive
 
+      - name: Cache Rust Build Cache
+        uses: actions/cache@v3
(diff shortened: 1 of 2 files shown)


tamird:
I manually re-ran the GH actions on this PR to observe the effect of the new cache; looks like we went from 29 minutes to just over 12 minutes. I'd say that's a nice win.

tamird:
Please hold off on merging this - I'm testing the effect of using rust-cache on rustc built from source.

tamird:
Removed the last commit; rust-cache does the wrong thing when used before the compiler is built, and because we're only building a stage 1 compiler, it seems to produce much bigger artifacts (1GB instead of ~300MB). This is now ready for review.

tamird:
Closing in favor of #51.

PR #51, opened 2023-06-02
github: use sccache-action

tamird:
Building rust from source takes our CI time from 3 minutes to 29
minutes. Cache the build products and pray that this reduces the time.

Limit the use of sccache to building rust to keep the cache size in
check; we already cache cargo artifacts separately.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -49,9 +49,16 @@ jobs:
           path: rust
           submodules: recursive
 
+      - name: Run sccache-cache
+        if: matrix.rust == 'nightly-from-source'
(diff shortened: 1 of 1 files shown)


tamird:
The cache-hit run here is about 2 minutes slower than in #50, but perhaps this is still better because it is more general. We'll need to build LLVM from source in #35, and sccache can help us there.

Another thing to explore is using the rust project's cache from S3. At a glance, it looks like sccache assumes it can always write to the cache (that's https://github.com/mozilla/sccache/issues/30), and as a result the rust project has access controls on their S3 buckets.

It appears there's no way to pay for more than 10GB of cache, but we can always do the same thing as the rust project and pay for storage in S3. I think we should land this over #50.

PR #52, opened 2023-06-02
Test on nightly-from-source only

tamird:
This continues from bbeeae567ae937a1d14602240e7ffbb607d03054; now that
https://github.com/rust-lang/rust/commit/b0ce4164f09be1221d004c4b84e49bdcbf973194
has made its way to the nightly channel, we can no longer use the host
target backdoor at all; all compiletest tests must run on a toolchain
that supports bpf.


alessandrod (approved):
😭

PR #68, opened 2023-07-06
Use rustc-build-sysroot to build sysroot

tamird:
This is much faster than building a stage 1 compiler.


alessandrod (approved):
This is amazing 😍

Issue #65, opened 2023-07-03
workflows: ensure bpf-linker works for aya integration tests


ajwerner:
It'd be nice to know that changes to bpf-linker don't break the aya integration tests.

Locally I find that a recent build of https://github.com/aya-rs/bpf-linker/pull/35 fails to link on many of the tests.

alessandrod:
https://github.com/aya-rs/aya/pull/634 introduces a dependency on the latest bpf-linker which includes this https://github.com/aya-rs/bpf-linker/pull/63

I agree it would be nice if we could somehow reify this dependency in code/something

PR #69, opened 2023-07-07
Run aya integration tests

tamird:
Fixes https://github.com/aya-rs/bpf-linker/issues/65.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -75,3 +75,37 @@ jobs:
       - name: Test
         if: matrix.rust == 'nightly'
         run: cargo hack test --feature-powerset
+
+      - uses: actions/checkout@v3
+        if: matrix.rust == 'nightly'
+        with:
+          repository: aya-rs/aya
+          path: aya
+
+      - uses: actions/checkout@v3
+        if: matrix.rust == 'nightly'
+        with:
+          repository: libbpf/libbpf
+          path: libbpf
+
+      - name: Install dependencies
+        if: matrix.rust == 'nightly'
+        # ubuntu-20.04 comes with clang prior to
+        # https://github.com/llvm/llvm-project/commit/6d6750696400 which does not implement
+        # __builtin_preserve_enum_value, used by relocation tests in aya.
+        #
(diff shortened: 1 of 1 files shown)


tamird:
@alessandrod @ajwerner this is ready for review (see https://github.com/aya-rs/aya/pull/641 for changes required to make this pass tests)

tamird:
(do not merge, this references https://github.com/aya-rs/aya/tree/logger-messages-plz until https://github.com/aya-rs/aya/pull/641 has been merged)

src/linker.rs
```diff
@@ -313,7 +313,7 @@ impl Linker {
         in_type: Option<InputType>,
     ) -> Result<(), LinkerError> {
         let mut data = Vec::new();
-        reader
+        let _: usize = reader
```
alessandrod, line 316: do we need the type annotation here?

src/llvm/mod.rs
```diff
@@ -220,7 +220,7 @@ pub unsafe fn optimize(
     }
 
     debug!("running function passes");
-    LLVMInitializeFunctionPassManager(fpm);
+    let _modified = LLVMInitializeFunctionPassManager(fpm) != 0;
```
alessandrod, line 223: what is the advantage of doing this? the compiler isn't complaining about it, we never check it, the code seems worse to read to me after the change

src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![deny(clippy::all)]
+#![deny(unused_results)]
```
alessandrod, line 2: Oh, this. But why? Did you run into some LLVM stuff that's supposed to be #[must_use]? If so let's annotate that instead?

tamird: Nah, this doesn't really belong in this PR specifically. It's a useful lint that avoids things like partial reads and writes which report the number of bytes and folks often forget to check; see `src/linker.rs` where I changed `read` to `read_exact` because the length read wasn't being checked -- this lint surfaced that.

I'll move this to another PR so we can discuss separately from this change.

alessandrod (approved):

PR #71, opened 2023-07-11
deny(unused_results)

tamird:
This reveals common mistakes such as neglecting to check for full reads.


src/llvm/mod.rs
```diff
@@ -220,7 +220,7 @@ pub unsafe fn optimize(
     }
 
     debug!("running function passes");
-    LLVMInitializeFunctionPassManager(fpm);
+    let _modified = LLVMInitializeFunctionPassManager(fpm) != 0;
```
dave-tucker, line 223: Could this `optimize` function return a Result instead?
Then rather than using `let _modified` everywhere, you could instead do something like this:
```rust
match LLVMInitializeFunctionPassManager(fpm) {
        ret @ 0.. => Ok(),
        ret => Err(ret),
}?;
```

tamird: The return value isn't an error. It tells you whether the operation modified the operand or not (it's an LLVMBool which happens to be an integer).

dave-tucker (approved):

PR #77, opened 2023-07-21
Add flag to treat LLVM errors as fatal

tamird:
This is apparently the contract we're supposed to uphold. Currently we sometimes miscompile things.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,3 +38,6 @@ rust-llvm = [
     "llvm-sys/disable-alltargets-init"
 ]
 default = ["rust-llvm"]
+
+[profile.release]
+debug = true
--- a/src/llvm/mod.rs
+++ b/src/llvm/mod.rs
@@ -309,17 +309,19 @@ pub unsafe fn internalize(
     }
 }
 
-pub extern "C" fn diagnostic_handler(info: LLVMDiagnosticInfoRef, _data: *mut c_void) {
-    let message = unsafe { CStr::from_ptr(LLVMGetDiagInfoDescription(info)) };
-    let message_s = message.to_str().unwrap();
+pub trait LLVMDiagnosticHandler {
+    fn handle_diagnostic(&mut self, severity: llvm_sys::LLVMDiagnosticSeverity, message: &str);
+}
 
-    use llvm_sys::LLVMDiagnosticSeverity::*;
-    match unsafe { LLVMGetDiagInfoSeverity(info) } {
-        LLVMDSError => error!("llvm: {}", message_s),
-        LLVMDSWarning => warn!("llvm: {}", message_s),
-        LLVMDSRemark => debug!("remark: {}", message_s),
-        LLVMDSNote => debug!("note: {}", message_s),
-    };
+pub extern "C" fn diagnostic_handler<T: LLVMDiagnosticHandler>(
+    info: LLVMDiagnosticInfoRef,
+    handler: *mut c_void,
+) {
+    let severity = unsafe { LLVMGetDiagInfoSeverity(info) };
+    let message = Message::from_ptr(unsafe { LLVMGetDiagInfoDescription(info) });
+    let handler = handler as *mut T;
+    unsafe { &mut *handler }
+        .handle_diagnostic(severity, message.as_c_str().unwrap().to_str().unwrap());
 }
 
 pub extern "C" fn fatal_error(reason: *const c_char) {
--- a/src/llvm/mod.rs
+++ b/src/llvm/mod.rs
@@ -74,9 +74,9 @@ pub unsafe fn find_embedded_bitcode(
     );
 
     let mut message = Message::new();
-    let bin = LLVMCreateBinary(buffer, ptr::null_mut(), message.as_mut_ptr());
+    let bin = LLVMCreateBinary(buffer, ptr::null_mut(), &mut *message);
     if bin.is_null() {
-        return Err(message.to_string());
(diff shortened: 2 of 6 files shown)


src/llvm/mod.rs
```diff
@@ -179,69 +181,130 @@ pub unsafe fn optimize(
...
+                return_type,
+                param_types.as_mut_ptr(),
+                param_count,
+                is_var_arg,
+            )
+        };
+
+        LLVMPositionBuilderBefore(builder, instruction);
+        // Call instructions can't have a name; attempting to set one crashes LLVM.
+        let name = [0];
```
ajwerner, line 295: nit: it means the same thing but might be more obvious if you wrote this `['\0']`

src/llvm/mod.rs
```diff
@@ -179,69 +181,130 @@ pub unsafe fn optimize(
...
+        // should return an error instead of panicking.
+        assert!(
+            !function.is_null(),
+            "missing mem intrinsic function {instruction_name}"
+        );
+        let function_type = LLVMGlobalGetValueType(function);
+        let function_type = {
+            let return_type = LLVMGetReturnType(function_type);
+            let param_count = LLVMCountParamTypes(function_type);
+            // Use the parameter types we're about to pass rather than the parameter types of the function.
```
ajwerner, line 277: Does this definitely work? Does LLVM do something smart to make sure the right data gets into the call?

src/llvm/mod.rs
```diff
@@ -179,69 +181,130 @@ pub unsafe fn optimize(
...
+    debug!("running optimization pass");
+    LLVMRunPasses(module, opt.as_ptr(), tm, options);
+
+    LLVMDisposePassBuilderOptions(options);
 
     // Some debug info generated by rustc seems to trigger a segfault in the
     // BTF code in llvm, so strip it until that is fixed
     LLVMStripModuleDebugInfo(module);
+
+    // Collect up all the memory intrinsics. We're going to replace these with
```
ajwerner, line 227: nit: this whole thing to rewrite the memory intrinsics feels like it could belong in its own function.

.github/workflows/build-test.yml
```diff
@@ -82,6 +82,7 @@ jobs:
           repository: aya-rs/aya
           path: aya
           submodules: recursive
+          ref: logs-wtf
```
ajwerner, line 85: I assume this is to be removed when https://github.com/aya-rs/aya/pull/683 merges, yeah?

ajwerner (approved):
LGTM

alessandrod:
I think adding the flag is good, but it should be disabled by default until the upstream LLVM conversation is resolved. I don't think we need to do the intrinsics => call change.

Cargo.toml
```diff
@@ -38,3 +38,9 @@ rust-llvm = [
     "llvm-sys/disable-alltargets-init"
 ]
 default = ["rust-llvm"]
+
+[profile.release]
+debug = true
+
+[patch.crates-io]
+aya-rustc-llvm-proxy = { git = "https://github.com/aya-rs/rustc-llvm-proxy.git" }
```
alessandrod, line 46: This needs to be removed before merging yeah?

tamird: You tell me - would you like to cut a release, or should we just take a git dep on the thing?

src/llvm/mod.rs
tamird: It could. What does it get us? I'm not a fan of indirection for its own sake.

src/llvm/mod.rs
tamird: I don't know how to answer the second half of this question.

Take a look at the commit view on this PR - you'll see that adding this makes CI go from failing to passing.

tamird:
> I think adding the flag is good, but it should be disabled by default until the upstream LLVM conversation is resolved.

What's the resolution you're referring to? It seems that there's no way to *not* treat these errors as fatal and still avoid miscompilation.

> I don't think we need to do the intrinsics => call change.

Are you suggesting we string match on that error and ignore it?

> The NewPM stuff should go in a separate PR.

Sure, it will. I wanted to demonstrate that the IR rewriting change prevents the miscompilation that otherwise happens.

src/llvm/mod.rs
ajwerner: Ack. I think it can improve readability in so far as a human can summarize what the function does. By having the logic to do this rewriting inline, it becomes more difficult to internalize the code structure from a single pass. I am not religious here.

src/llvm/mod.rs
ajwerner: Well, we had a value that was one byte being passed to the intrinsic, and now for the builtin we're passing a value that is 4 bytes, but we're saying its type is one byte (right?). Are we just seeing it work because we're running on little-endian machines and we're passing the arguments through registers such that a 4-byte value is the same as a 1-byte value?

tamird: Added an explicit cast.

ajwerner: The code is cleaner this way, even if it was all correct without the cast. Thank you.

src/llvm/mod.rs
tamird: This should be more obvious now.

Cargo.toml
tamird: Pulled this out of this PR.

ajwerner (approved):
LGTM

src/llvm/mod.rs
tamird: This code is no longer in this PR.

PR #78, opened 2023-07-24
Missing libpolly-16-dev dep for apt.llvm.org repo instructions

ukreator:
Minor README fix `cargo install bpf-linker --no-default-features` fails to build if libpolly-16-dev is not installed

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ cargo install bpf-linker
 
 On Debian based distributions you need to install the `llvm-16-dev` and
 `libclang-16-dev` packages. If your distro doesn't have them you can get them
-from the official LLVM repo at https://apt.llvm.org.
+from the official LLVM repo at https://apt.llvm.org. You need to install libpolly-16-dev too if you use this repo.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,7 +32,7 @@ cargo install bpf-linker
 
 On Debian based distributions you need to install the `llvm-16-dev` and
 `libclang-16-dev` packages. If your distro doesn't have them you can get them
-from the official LLVM repo at https://apt.llvm.org.
+from the official LLVM repo at https://apt.llvm.org. You need to install libpolly-16-dev too if you use this repo.
```
tamird, line 35: why is this not in the first sentence that mentions llvm-16-dev and libclang-16-dev?

also please keep the column width consistent.

tamird (changes requested):

README.md
ukreator: my impression was it's only needed for packages from apt.llvm.org. Double checked with the recent Ubuntu - it's needed there too. Will fix my changes shortly

tamird (approved):

PR #87, opened 2023-08-07
fixed lto flag in BTF tests

qjerome:
There was lto=true inside BTF test, which hides some bugs encountered in the linker.
In order to be able to remove `lto=true` in BTF tests, a new sysroot built with DI needs to be used. The reason for this is a bug in LLVM which prevents to have code with DI and code without DI built together.

NB: we cannot use a single sysroot built with DI for all the tests (BTF + assembly) because assembly tests are crashing the linker when built with DI.


qjerome:
Assembly test crashing with DI sysroot tracked in: https://github.com/aya-rs/bpf-linker/issues/88

PR #92, opened 2023-08-08
github: add CI for feature branches

tamird:
See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#using-filters.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,9 +4,11 @@ on:
   push:
     branches:
       - main
+      - 'feature/**'
   pull_request:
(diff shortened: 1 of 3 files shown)


vadorovsky (approved):

tamird:
this won't pass CI until https://github.com/aya-rs/aya/pull/725 lands.

tamird:
Turns out https://github.com/aya-rs/aya/pull/725 wasn't enough, so we'll need https://github.com/aya-rs/aya/pull/733 so we can run the integration tests on a VM with a newer kernel.

PR #93, opened 2023-08-08
Add test for ilog2

addisoncrump:
It seems that LLVM does not have an instruction pattern for ilog2/leading bits/trailing bits. This adds a test case for this behaviour.


tests/assembly/ilog2.rs
```diff
@@ -0,0 +1,13 @@
+// assembly-output: bpf-linker
+// compile-flags: --crate-type cdylib -C link-arg=--unroll-loops -C link-arg=-O3
+#![no_std]
+// bpf target did not historically support the ilog2 operation because LLVM did
```
tamird, line 4: what are you testing here? There's no CHECK; can you add some?

tests/assembly/ilog2.rs
```diff
@@ -0,0 +1,13 @@
+// assembly-output: bpf-linker
+// compile-flags: --crate-type cdylib -C link-arg=--unroll-loops -C link-arg=-O3
```
tamird, line 2: do we need unroll-loops and 03? seems superfluous, see https://en.wikipedia.org/wiki/Chekhov%27s_gun

tamird (changes requested):

tests/assembly/ilog2.rs
addisoncrump: TBH I do not remember the details of how to use this testing. To my memory, I wanted this to build successfully, but it currently doesn't because `ilog2` IL is refused by LLVM.

tests/assembly/ilog2.rs
addisoncrump: Almost certainly copy-pasted this :grinning:

tests/assembly/ilog2.rs
tamird: OK. Should we close this?

addisoncrump: No, because this is actually affecting some of my code. Right now, I have to manually compute ilog2. If I ever depended on a crate which uses this functionality, I would be screwed :sweat_smile:

addisoncrump: How do I correctly write this testcase to check that the compilation is successful?

tamird: What is it you're looking for, here?

tamird: > How do I correctly write this testcase to check that the compilation is successful?

Read the llvm-filecheck docs? Have you tried running this locally?

addisoncrump: > What is it you're looking for, here?

Historically, ilog2 would cause a linkage error that was very difficult to diagnose.

> Have you tried running this locally?

Not recently, no. I'll try to take a look at it this week and update it according to the llvm-filecheck docs.

tamird:
@addisoncrump you planning to come back to this?

PR #99, opened 2023-08-11
Fix tests

qjerome:
Build tests sysroot with DI:
* fixed ignore-inline-never.rs that is failing because of FileCheck matching DI instead of expected check
* removed lto=true from BTF tests

NB: this requires patched LLVM https://github.com/rust-lang/llvm-project/commit/d06f0cd0977dd16f09c259133f36c7e8b9f0d322


qjerome:
Dammit the linker is crashing !

tamird:
I think you might need https://github.com/0xrawsec/bpf-linker/commit/922a95d2cc855a415e49f05098b0364b47bfeb9a.

qjerome:
IDK, only testing will tell me if that solves the issue. However, investigating that issue will have to wait one week as I'll not be able to work on Aya next week !  I deliberately chose to split my PRs as I understood @alessandrod prefers small PR addressing one problem at a time.

qjerome:
The crash occurred when hitting a DIFlagFwdDecl. This get fixed by applying a commit I wanted to address in a separate PR.

qjerome:
Test passed but "Download debian kernels" step failed ! What can I do ?
cc: @alessandrod @tamird

tamird:
You need to rebase.

qjerome:
How do you guys usually resolve merge conflicts in Cargo.lock file when merging/rebasing ?

tests/assembly/ignore-inline-never.rs
```diff
@@ -21,6 +21,6 @@ fn actually_inlined(a: u64) -> u64 {
 pub extern "C" fn fun(a: u64) -> u64 {
     // CHECK-LABEL: fun:
     actually_inlined(a)
-    // CHECK-NEXT: r{{[0-9]}} = r{{[0-9]}}
```
tamird, line 24: Can you explain this change in a commit message?

tests/tests.rs
```diff
@@ -72,6 +72,7 @@ fn compile_test() {
     let () = rustc_build_sysroot::SysrootBuilder::new(&directory, target)
         .build_mode(rustc_build_sysroot::BuildMode::Build)
         .sysroot_config(rustc_build_sysroot::SysrootConfig::NoStd)
+        .rustflag("-Cdebuginfo=2")
```
tamird, line 75: can you add a comment here to explain this? can we rely on the sysroot shipped with rustup to have DI?

src/llvm/di.rs
```diff
@@ -80,7 +80,19 @@ impl DIFix {
 
                         // we detect this is a variadic enum if the child element is a DW_TAG_variant_part
                         let elements = LLVMGetOperand(value, 4);
```
tamird, line 82: this call is pointless if `metadata` is a forward declaration, right?

tamird (commented):
Can you squash this into a single commit? As far as I can tell all the changes are required to pass tests; none of these commits make sense on their own.

src/llvm/di.rs
qjerome: Yes it is pointless :). 
Removing this requires changing a bit the logic but I can do it

tests/assembly/ignore-inline-never.rs
qjerome: Will do

qjerome:
Fixed according your comments

src/llvm/di.rs
```diff
@@ -78,45 +78,51 @@ impl DIFix {
...
-                                            return None;
-                                        }
-                                        let m = LLVMValueAsMetadata(v);
-                                        let metadata_kind = LLVMGetMetadataKind(m);
-                                        let file_operand_index = match metadata_kind {
+                        let flags = LLVMDITypeGetFlags(metadata);
+
+                        // LLVMGetOperand(value, 4) returns NULL only for DI with flags: DIFlagFwdDecl
+                        // declarations with DIFlagFwdDecl always get declared elsewhere without this flag
+                        if flags != LLVMDIFlagFwdDecl {
```
tamird, line 85: can't we just return if flags contains LLVMDIFlagFwdDecl?

tests/tests.rs
tamird: You'll probably see it when you build some BPF crate using `-Z build-std=core`; if it builds without DI, we might have a problem.

The fact that you're using `cargo test` doesn't mean anything because the tests are building the sysroot (right here).

alessandrod: `-Zbuild-std=core` will build core, so it'll generate DI if DI `-C debuginfo=2` is passed

src/llvm/di.rs
```diff
@@ -78,10 +78,19 @@ impl DIFix {
                         // variadic enum not supported => emit warning and strip out the children array
                         // i.e. pub enum Foo { Bar, Baz(u32), Bad(u64, u64) }
 
+                        let flags = LLVMDITypeGetFlags(metadata);
+
+                        // LLVMGetOperand(value, 4) returns NULL only for DI with flags: DIFlagFwdDecl
```
alessandrod, line 83: This comment should say:

```
This is a forward declaration. We don't need to do anything on the declaration, we're going to process the actual definition.
```

alessandrod:
lgtm! See comment nit.

Also can you update the title to say "build tests sysroot with DI; skip forward declaration when massaging DI". In the future please send separate PRs for unrelated changes (DI for sysroot and skipping declarations don't seem to be related)

qjerome:
@alessandrod thanks for your review
Just FYI because I guess you don't have the time to follow all PRs. My initial plan was indeed to make a separate PR but I had to integrate "skipping declarations" patch here because test with nightly rust were failing without it (see https://github.com/aya-rs/bpf-linker/pull/99#issuecomment-1674441317).

Issue #123, opened 2023-09-26
LLVM17 cold callsite categorization breaks inlining


qjerome:
This is a wrap up of what has been discussed in: https://discord.com/channels/855676609003651072/1026937450624450652/1155989596870549584

After migration to LLVM17, LLVM has the functionality to track cold and hot callsite.
This categorization, based on call frequency,  impacts the inlining strategy in the following way:
1. cold sites will less likely be inlined 
2. cold site status superseed `inlinehint` LLVM attribute set by `#[inline]` tag in Rust
3. hot call sites will be more likely to get inlined

While the strategy concerning hot callsite does not seem to be an issue, the one concerning cold ones seems problematic.
Even though, `#[inline(always)]` never guarantee that the function will get inlined, it seems to happen rather frequently that `always_inline` functions **also categorized as cold callsite** do not get inlined at all. This creates loads of errors in `bpf-linker` preventing the linking from happening. All the errors observed being due to too large return values (i.e. > 64 bits in size).

**Why this change is needed:** there must be a way to programmatically ensure that functions that return results larger than 64 bits get inlined https://discord.com/channels/855676609003651072/1026937450624450652/1155834661558288424

The best solution to opt for and have an inlining strategy close to what bpf-linker had with LLVM16 is to disable cold callsite categorization. This is possible by passing `--llvm-args=--cold-callsite-rel-freq=0` to `bpf-linker`.

PR #124, opened 2023-09-27
cold callsite disabled

qjerome:
fix https://github.com/aya-rs/bpf-linker/issues/123

I am sorry, I did not managed to create a tests with functions being categorized as cold callsites.
However, I did test this fix by successfully building and running https://github.com/0xrawsec/kunai on several LTS kernels.
Without this fix linking was failing because lots of `#[inline(always)]` functions didn't get inlined as it used to be with LLVM16. All this is explained in the bpf-linker issue referenced above.

### DIFF
--- a/src/linker.rs
+++ b/src/linker.rs
@@ -482,6 +482,13 @@ impl Linker {
     fn llvm_init(&mut self) {
         let mut args = Vec::<Cow<str>>::new();
         args.push("bpf-linker".into());
+        // Disable cold call site detection. Many accessors in aya-ebpf return Result<T, E>
+        // where the layout is larger than 64 bits, but the LLVM BPF target only supports
+        // up to 64 bits return values. Since the accessors are tiny in terms of code, we
+        // avoid the issue by annotating them with #[inline(always)]. If they are classified
+        // as cold though - and they often are starting from LLVM17 - #[inline(always)]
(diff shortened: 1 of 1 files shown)


alessandrod:
Context why this change is needed: there must be a way to programmatically ensure that functions that return results larger than 64 bits get inlined https://discord.com/channels/855676609003651072/1026937450624450652/1155834661558288424

src/linker.rs
```diff
@@ -482,6 +482,8 @@ impl Linker {
     fn llvm_init(&mut self) {
         let mut args = Vec::<Cow<str>>::new();
         args.push("bpf-linker".into());
+        // starting from LLVM17 we need to disable cold callsite to avoid inlining issues
```
alessandrod, line 485: Can you please change the comment to:

```
Disable cold call site detection. Many accessors in aya-ebpf return Result<T, E> where the layout is
larger than 64 bits, but the LLVM BPF target only supports up to 64 bits return values. Since the
accessors are tiny in terms of code, we avoid the issue by annotating them with #[inline(always)].
If they are classified as cold though - and they often are starting from LLVM17 - #[inline(always)]
is ignored and the BPF target fails codegen. 
```

qjerome: change applied

alessandrod (approved):

PR #140, opened 2023-10-22
Cargo.toml: remove aya-rustc-llvm-proxy feature

tamird:
This doesn't make sense as a standalone feature.


vadorovsky (approved):

tamird:
Merging without waiting for aya integration tests because this fixes a linking failure.

PR #190, opened 2024-03-21
bump rustc-llvm-proxy

alessandrod:


tamird:
This is not enough, you need to bump it in the lock (this is a binary, not a library).

PR #197, opened 2024-04-19
Log errors by default

tamird:
Remove imperative `clap::Error::exit`.

### DIFF
--- a/src/bin/bpf-linker.rs
+++ b/src/bin/bpf-linker.rs
@@ -3,13 +3,19 @@
 #[cfg(feature = "rust-llvm")]
 extern crate aya_rustc_llvm_proxy;
 
-use std::{env, fs, io, path::PathBuf, str::FromStr};
+use std::{
+    env, fs, io,
+    path::{Component, Path, PathBuf},
+    str::FromStr,
+};
 
 use bpf_linker::{Cpu, Linker, LinkerOptions, OptLevel, OutputType};
-use clap::Parser;
+use clap::{
+    builder::{PathBufValueParser, TypedValueParser as _},
+    Parser,
+};
 use thiserror::Error;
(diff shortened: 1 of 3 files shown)


tamird:
@vadorovsky looks like eff4440f73aa380eba658a5b5a93f5fe4e298a2e accidentally lost logging if an explicit level was not passed to the CLI.

vadorovsky (commented):
Awesome, happy to see the custom `error` function to be gone and replaced with anyhow! So far it looks good, I'm just going to try it a bit before approving.

src/bin/bpf-linker.rs
```diff
@@ -192,57 +216,40 @@ fn main() {
...
                 let subscriber = subscriber_registry.with(tracing_layer(io::stderr));
-                (Box::new(subscriber), None)
+                (either::Right(subscriber), None)
             }
         };
-        if let Err(e) = tracing::subscriber::set_global_default(subscriber) {
-            error(&e.to_string(), clap::error::ErrorKind::Format);
-        }
-    }
+        either::for_both!(subscriber, s => tracing::subscriber::set_global_default(s))?;
```
alessandrod, line 243: unless it's already in the dep tree, it doesn't seem worth adding a new dep just for this?

alessandrod (commented):
lgtm but let's let @vadorovsky try what he wants to try.

As per comment I wouldn't add a new dep tho.

src/bin/bpf-linker.rs
tamird: Take a look at Cargo.lock; it's already in there.

alessandrod: It's a dev dependency right?

tamird: Good point, removed.

PR #208, opened 2024-07-13
Support --version flag

ymgyt:
I thought it would be convenient to display the version of bpf-linker, so I added support for the `--version` flag.

When you run bpf-linker --version (or -V), it displays:
```
bpf-linker 0.9.12
```

In this implementation, I ensured that when --help or --version is specified, the command returns a successful status code

### DIFF
--- a/src/bin/bpf-linker.rs
+++ b/src/bin/bpf-linker.rs
@@ -12,6 +12,7 @@ use std::{
 use bpf_linker::{Cpu, Linker, LinkerOptions, OptLevel, OutputType};
 use clap::{
     builder::{PathBufValueParser, TypedValueParser as _},
+    error::ErrorKind,
     Parser,
 };
 use thiserror::Error;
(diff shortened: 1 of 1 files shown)


src/bin/bpf-linker.rs
```diff
@@ -216,7 +218,16 @@ fn main() -> anyhow::Result<()> {
         export,
         fatal_errors,
         _debug,
-    } = Parser::try_parse_from(args)?;
+    } = match Parser::try_parse_from(args) {
+        Ok(command_line) => command_line,
+        Err(err) => match err.kind() {
+            ErrorKind::DisplayHelp | ErrorKind::DisplayVersion => {
+                print!("{err}");
+                return Ok(());
```
ymgyt, line 226: If an error were returned as is, it would display as:
```
Error: bpf-linker 0.9.12
```

Therefore, I made sure to return `Ok` instead.

alessandrod (approved):

alessandrod:
Thank you!

PR #260, opened 2025-02-26
fix: c_void enum in BTF

qjerome:
This PR fixes an issue with kernels < 5.4 failing to load BTF because of the [`c_void` enum](https://doc.rust-lang.org/std/ffi/enum.c_void.html).

When used in BPF code, the Rust `c_void` type (actually an enum) generates the following BTF:

```
#17: <ENUM> 'c_void' sz:1 n:2
  #00 __variant1 = 0
  #01 __variant2 = 1
```

It appears that kernels < 5.4 do not allow BTF enums with a size of 1 byte.

The fix consists of replacing the `c_void` enum MDNode with `i8` DI BasicType information.

To achieve this, a small refactoring of `DISanitizer::visit_mdnode` was necessary. The old `visit_mdnode` had only one `MDNode` argument, which would have required an additional argument to pass the `Item::Operand` to fix the MDNode. This was not optimal because both the `Operand` and the `MDNode` carry the same `LLVMValueRef`, which I found confusing (two arguments with the same information). The refactoring aims to make things cleaner and less confusing.

Another issue arose when using LLVM compiled with debug assertions. A casting assertion failed at [this line](https://github.com/llvm/llvm-project/blob/3e59710604e84d563962f259fef8b8a18c23f04e/llvm/lib/CodeGen/AsmPrinter/DwarfDebug.cpp#L1218):

```cpp
1218    for (auto *Ty : CUNode->getEnumTypes())
1219      CU.getOrCreateTypeDIE(cast<DIType>(Ty));
1220
```

The exact assertion failing downstream is `llvm::cast_if_present<llvm::DICompositeType, llvm::MDOperand> (Val=...)`. By inspecting `Val` (with gdb), we observe the following:

1. Its `MetadataKind` is `0xc` -> DIDerivedType
   ```
   print ((llvm::Metadata*) Val).SubclassID
   $16 = 0xc
   ```

2. This DIDerivedType is actually a pointer to the `DIBaseType` we replaced the `c_void` enum with:
   ```
   print ((llvm::DIDerivedType*) Val).DWARFAddressSpace
   $17 = std::optional = {
     [contained value] = 0x5
   }
   ```

   Note: `0x5` is the LLVMDWARFTypeEncoding we use in the `i8` DIBasicType to replace the `c_void` enum. I have confirmed this value corresponds to the LLVMDWARFTypeEncoding value.

I was unable to fully understand why and how this replacement happens. I speculate it is a side effect of replacing the `c_void` DICompositeType using `LLVMReplaceMDNodeOperandWith`. One observation is that it tends to happen when we replace the `c_void` enum use in an [MDTuple](https://llvm.org/doxygen/classllvm_1_1MDTuple.html).

However, the fact remains that the replaced `c_void` enums no longer need to exist in the DICompileUnit, as we have replaced any use of it. The proposed solution is to delete it completely using the `DISanitizer::fix_di_compile_units` function.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/260)

### DIFF
--- a/src/llvm/di.rs
+++ b/src/llvm/di.rs
@@ -6,12 +6,14 @@ use std::{
     ptr,
 };
 
-use gimli::{DW_TAG_pointer_type, DW_TAG_structure_type, DW_TAG_variant_part};
+use gimli::{
+    DW_TAG_enumeration_type, DW_TAG_pointer_type, DW_TAG_structure_type, DW_TAG_variant_part,
+};
 use llvm_sys::{core::*, debuginfo::*, prelude::*};
 use tracing::{span, trace, warn, Level};
 
 use super::types::{
-    di::DIType,
+    di::{DIBasicType, DIBasicTypeKind, DICompileUnit, DIType},
     ir::{Function, MDNode, Metadata, Value},
 };
 use crate::llvm::{iter::*, types::di::DISubprogram};
@@ -21,13 +23,14 @@ use crate::llvm::{iter::*, types::di::DISubprogram};
 // backward compatibility
 const MAX_KSYM_NAME_LEN: usize = 128;
 
-pub struct DISanitizer {
+pub struct DISanitizer<'ctx> {
     context: LLVMContextRef,
     module: LLVMModuleRef,
     builder: LLVMDIBuilderRef,
     visited_nodes: HashSet<u64>,
     replace_operands: HashMap<u64, LLVMMetadataRef>,
     skipped_types: Vec<String>,
+    basic_types: HashMap<DIBasicTypeKind, DIBasicType<'ctx>>,
 }
 
 // Sanitize Rust type names to be valid C type names.
@@ -58,8 +61,13 @@ fn sanitize_type_name<T: AsRef<str>>(name: T) -> String {
     n
 }
 
-impl DISanitizer {
-    pub fn new(context: LLVMContextRef, module: LLVMModuleRef) -> DISanitizer {
+/// [`LLVMGetMetadataKind`] wrapper
+fn kind(m: LLVMMetadataRef) -> LLVMMetadataKind {
+    unsafe { LLVMGetMetadataKind(m) }
+}
+
+impl<'ctx> DISanitizer<'_> {
+    pub fn new(context: LLVMContextRef, module: LLVMModuleRef) -> DISanitizer<'ctx> {
         DISanitizer {
             context,
             module,
@@ -67,15 +75,42 @@ impl DISanitizer {
             visited_nodes: HashSet::new(),
             replace_operands: HashMap::new(),
             skipped_types: Vec::new(),
+            basic_types: HashMap::new(),
         }
     }
 
-    fn visit_mdnode(&mut self, mdnode: MDNode) {
+    /// Returns a [`DIBasicType`] given a [`DIBasicTypeKind`].
+    fn di_basic_type(&'ctx mut self, di_bt: DIBasicTypeKind) -> &'ctx DIBasicType<'ctx> {
+        self.basic_types
+            .entry(di_bt)
+            .or_insert_with(|| DIBasicType::llvm_create(self.context, self.builder, di_bt))
+    }
+
+    fn visit_mdnode_item(&mut self, item: &mut Item) {
+        let Some(mdnode) = item.as_mdnode() else {
+            return;
+        };
+
         match mdnode.try_into().expect("MDNode is not Metadata") {
             Metadata::DICompositeType(mut di_composite_type) => {
                 #[allow(clippy::single_match)]
                 #[allow(non_upper_case_globals)]
                 match di_composite_type.tag() {
+                    DW_TAG_enumeration_type => {
+                        if let Some(name) = di_composite_type.name() {
+                            // we found the c_void enum
+                            if name == c"c_void" && di_composite_type.size_in_bits() == 8 {
+                                if let Item::Operand(ref mut op) = item {
+                                    // get i8 DIBasicType
+                                    let i8_bt = self.di_basic_type(DIBasicTypeKind::I8);
+                                    op.replace(i8_bt.value_ref);
+                                } else {
+                                    // c_void enum is not an Item::Operand so we cannot replace it
+                                    warn!("failed at replacing c_void enum, it might result in BTF parsing errors in kernels < 5.4")
+                                }
+                            }
+                        }
+                    }
                     DW_TAG_structure_type => {
                         let names = match di_composite_type.name() {
                             Some(name) => {
@@ -247,9 +282,7 @@ impl DISanitizer {
             return;
         }
 
-        if let Value::MDNode(mdnode) = value.clone() {
-            self.visit_mdnode(mdnode)
-        }
+        self.visit_mdnode_item(&mut item);
 
         if let Some(operands) = value.operands() {
             for (index, operand) in operands.enumerate() {
@@ -305,6 +338,8 @@ impl DISanitizer {
                 self.skipped_types.join(", ")
             );
         }
+
+        self.fix_di_compile_units();
 
         unsafe { LLVMDisposeDIBuilder(self.builder) };
     }
@@ -404,6 +439,58 @@ impl DISanitizer {
 
         replace
     }
+
+    fn di_compile_units(&self) -> Vec<DICompileUnit> {
+        let compile_unit_name = c"llvm.dbg.cu";
+
+        // Get the number of DICompileUnit.
+        let num_di_cu =
+            unsafe { LLVMGetNamedMetadataNumOperands(self.module, compile_unit_name.as_ptr()) };
+
+        // Create a vector to hold them all.
+        let mut di_cus: Vec<LLVMValueRef> = vec![core::ptr::null_mut(); num_di_cu as usize];
+
+        unsafe {
+            LLVMGetNamedMetadataOperands(
+                self.module,
+                compile_unit_name.as_ptr(),
+                di_cus.as_mut_ptr(),
+            )
+        };
+
+        di_cus
+            .into_iter()
+            .map(|v| unsafe { DICompileUnit::from_value_ref(v) })
+            .collect()
+    }
+
+    /// Removes replaced `c_void` Rust enum from all [`DICompileUnit`]. After
+    /// replacing `c_void` enum with a [`DIBasicType`] the [`DICompileUnit`] still
+    /// hold a reference to the enum and still believes it is a [`DICompositeType`]
+    /// which triggers a casting assertion in LLVM.
+    fn fix_di_compile_units(&mut self) {
+        for mut di_cu in self.di_compile_units() {
+            let tmp_cu = di_cu.clone();
+            let need_replace = tmp_cu.enum_types().any(|ct| {
+                matches!(
+                    kind(ct.metadata_ref),
+                    LLVMMetadataKind::LLVMDIBasicTypeMetadataKind
+                )
+            });
+
+            if need_replace {
+                di_cu.replace_enum_types(
+                    self.builder,
+                    tmp_cu.enum_types().filter(|ct| {
+                        !matches!(
+                            kind(ct.metadata_ref),
+                            LLVMMetadataKind::LLVMDIBasicTypeMetadataKind
+                        )
+                    }),
+                );
+            }
+        }
+    }
 }
 
 #[derive(Clone, Debug, Eq, PartialEq)]
@@ -438,6 +525,16 @@ impl Operand {
 }
 
 impl Item {
+    /// Returns the [`Item`] as [`MDNode`] only if [`Item::is_mdnode`] is `true` else `None`.
+    fn as_mdnode(&self) -> Option<MDNode<'_>> {
+        let is_mdnode = unsafe { !LLVMIsAMDNode(self.value_ref()).is_null() };
+        if is_mdnode {
+            Some(unsafe { MDNode::from_value_ref(self.value_ref()) })
+        } else {
+            None
+        }
+    }
+
     fn value_ref(&self) -> LLVMValueRef {
         match self {
             Item::GlobalVariable(value)
--- a/src/llvm/types/di.rs
+++ b/src/llvm/types/di.rs
@@ -7,13 +7,17 @@ use std::{
 
 use gimli::DwTag;
 use llvm_sys::{
-    core::{LLVMGetNumOperands, LLVMGetOperand, LLVMReplaceMDNodeOperandWith, LLVMValueAsMetadata},
-    debuginfo::{
-        LLVMDIFileGetFilename, LLVMDIFlags, LLVMDIScopeGetFile, LLVMDISubprogramGetLine,
-        LLVMDITypeGetFlags, LLVMDITypeGetLine, LLVMDITypeGetName, LLVMDITypeGetOffsetInBits,
-        LLVMGetDINodeTag,
+    core::{
+        LLVMGetNumOperands, LLVMGetOperand, LLVMMetadataAsValue, LLVMReplaceMDNodeOperandWith,
+        LLVMValueAsMetadata,
     },
-    prelude::{LLVMContextRef, LLVMMetadataRef, LLVMValueRef},
+    debuginfo::{
+        LLVMDIBuilderCreateBasicType, LLVMDIBuilderGetOrCreateTypeArray, LLVMDIFileGetFilename,
+        LLVMDIFlags, LLVMDIScopeGetFile, LLVMDISubprogramGetLine, LLVMDITypeGetFlags,
+        LLVMDITypeGetLine, LLVMDITypeGetName, LLVMDITypeGetOffsetInBits, LLVMDITypeGetSizeInBits,
+        LLVMDWARFTypeEncoding, LLVMGetDINodeTag,
+    },
+    prelude::{LLVMContextRef, LLVMDIBuilderRef, LLVMMetadataRef, LLVMValueRef},
 };
 
 use crate::llvm::{
@@ -221,7 +225,7 @@ enum DICompositeTypeOperand {
 /// Composite type is a kind of type that can include other types, such as
 /// structures, enums, unions, etc.
 pub struct DICompositeType<'ctx> {
-    metadata_ref: LLVMMetadataRef,
+    pub(crate) metadata_ref: LLVMMetadataRef,
     value_ref: LLVMValueRef,
     _marker: PhantomData<&'ctx ()>,
 }
@@ -307,6 +311,11 @@ impl DICompositeType<'_> {
     /// Returns a DWARF tag of the given composite type.
     pub fn tag(&self) -> DwTag {
         unsafe { di_node_tag(self.metadata_ref) }
+    }
+
+    /// Returns the size in bits of the composite type.
+    pub fn size_in_bits(&self) -> u64 {
+        unsafe { LLVMDITypeGetSizeInBits(LLVMValueAsMetadata(self.value_ref)) }
     }
 }
 
@@ -428,5 +437,141 @@ impl DISubprogram<'_> {
                 nodes,
             )
         };
+    }
+}
+
+/// Represents the operands for a [`DICompileUnit`]. The enum values correspond
+/// to the operand indices within metadata nodes.
+#[repr(u32)]
+enum DICompileUnitOperand {
+    EnumTypes = 4,
+}
+
+/// Represents the debug information for a compile unit in LLVM IR.
+#[derive(Clone)]
+pub struct DICompileUnit<'ctx> {
+    value_ref: LLVMValueRef,
+    _marker: PhantomData<&'ctx ()>,
+}
+
+impl<'ctx> DICompileUnit<'ctx> {
+    /// Constructs a new [`DICompileUnit`] from the given `value_ref`.
+    ///
+    /// # Safety
+    ///
+    /// This method assumes that the provided `value_ref` corresponds to a valid
+    /// instance of [LLVM `DICompileUnit`](https://llvm.org/doxygen/classllvm_1_1DICompileUnit.html).
+    /// It's the caller's responsibility to ensure this invariant, as this
+    /// method doesn't perform any valiation checks.
+    pub(crate) unsafe fn from_value_ref(value_ref: LLVMValueRef) -> Self {
+        Self {
+            value_ref,
+            _marker: PhantomData,
+        }
+    }
+
+    pub fn enum_types(&self) -> impl Iterator<Item = DICompositeType> {
+        let llvm_enum_types =
+            unsafe { LLVMGetOperand(self.value_ref, DICompileUnitOperand::EnumTypes as u32) };
+
+        let llvm_enum_types_len = if llvm_enum_types.is_null() {
+            0
+        } else {
+            unsafe { LLVMGetNumOperands(llvm_enum_types) }
+        };
+
+        (0..llvm_enum_types_len).map(move |i| unsafe {
+            let enum_type = LLVMGetOperand(llvm_enum_types, i as u32);
+            DICompositeType::from_value_ref(enum_type)
+        })
+    }
+
+    pub fn replace_enum_types<I>(&mut self, builder: LLVMDIBuilderRef, rep: I)
+    where
+        I: IntoIterator<Item = DICompositeType<'ctx>>,
+    {
+        let mut rep: Vec<_> = rep.into_iter().map(|dct| dct.metadata_ref).collect();
+
+        unsafe {
+            let enum_array =
+                LLVMDIBuilderGetOrCreateTypeArray(builder, rep.as_mut_ptr(), rep.len());
+            LLVMReplaceMDNodeOperandWith(
+                self.value_ref,
+                DICompileUnitOperand::EnumTypes as u32,
+                enum_array,
+            );
+        }
+    }
+}
+
+/// Represents [`DIBasicType`] kinds.
+#[derive(Hash, PartialEq, Eq, Clone, Copy)]
+pub enum DIBasicTypeKind {
+    I8,
+}
+
+impl DIBasicTypeKind {
+    fn name(&self) -> &'static str {
+        match self {
+            Self::I8 => "i8",
+        }
+    }
+
+    fn size_in_bits(&self) -> u64 {
+        match self {
+            Self::I8 => 8,
+        }
+    }
+
+    // DWARF encoding https://llvm.org/docs/LangRef.html#dibasictype
+    fn dwarf_type_encoding(&self) -> LLVMDWARFTypeEncoding {
+        match self {
+            // DW_ATE_signed
+            Self::I8 => gimli::DW_ATE_signed.0.into(),
+        }
+    }
+}
+
+/// Represents the debug information for a basic type in LLVM IR.
+pub struct DIBasicType<'ctx> {
+    pub(crate) value_ref: LLVMValueRef,
+    _marker: PhantomData<&'ctx ()>,
+}
+
+impl DIBasicType<'_> {
+    /// Constructs a new [`DIBasicType`] from the given `value_ref`.
+    ///
+    /// # Safety
+    ///
+    /// This method assumes that the provided `value_ref` corresponds to a valid
+    /// instance of [LLVM `DIBasicType`](https://llvm.org/doxygen/classllvm_1_1DIBasicType.html).
+    /// It's the caller's responsibility to ensure this invariant, as this
+    /// method doesn't perform any valiation checks.
+    pub(crate) unsafe fn from_value_ref(value_ref: LLVMValueRef) -> Self {
+        Self {
+            value_ref,
+            _marker: PhantomData,
+        }
+    }
+
+    /// Creates a new [`DIBasicType`] of `kind` given a `context` and a `builder`.
+    pub fn llvm_create(
+        ctx: LLVMContextRef,
+        builder: LLVMDIBuilderRef,
+        kind: DIBasicTypeKind,
+    ) -> Self {
+        let name = kind.name();
+        let metadata_ref = unsafe {
+            LLVMDIBuilderCreateBasicType(
+                builder,
+                name.as_ptr() as *const _,
+                name.len(),
+                kind.size_in_bits(),
+                kind.dwarf_type_encoding(),
+                0,
+            )
+        };
+
+        unsafe { Self::from_value_ref(LLVMMetadataAsValue(ctx, metadata_ref)) }
     }
 }
--- a/src/llvm/types/ir.rs
+++ b/src/llvm/types/ir.rs
@@ -22,7 +22,7 @@ use llvm_sys::{
 use crate::llvm::{
     iter::IterBasicBlocks as _,
     symbol_name,
-    types::di::{DICompositeType, DIDerivedType, DISubprogram, DIType},
+    types::di::{DICompileUnit, DICompositeType, DIDerivedType, DISubprogram, DIType},
     Message,
 };
 
@@ -112,6 +112,7 @@ pub enum Metadata<'ctx> {
     DICompositeType(DICompositeType<'ctx>),
     DIDerivedType(DIDerivedType<'ctx>),
     DISubprogram(DISubprogram<'ctx>),
+    DICompileUnit(#[allow(dead_code)] DICompileUnit<'ctx>),
     Other(#[allow(dead_code)] LLVMValueRef),
 }
 
@@ -140,6 +141,10 @@ impl Metadata<'_> {
                 let di_subprogram = unsafe { DISubprogram::from_value_ref(value) };
                 Metadata::DISubprogram(di_subprogram)
             }
+            LLVMMetadataKind::LLVMDICompileUnitMetadataKind => {
+                let di_compile_unit = unsafe { DICompileUnit::from_value_ref(value) };
+                Metadata::DICompileUnit(di_compile_unit)
+            }
             LLVMMetadataKind::LLVMDIGlobalVariableMetadataKind
             | LLVMMetadataKind::LLVMDICommonBlockMetadataKind
             | LLVMMetadataKind::LLVMMDStringMetadataKind
@@ -156,7 +161,6 @@ impl Metadata<'_> {
             | LLVMMetadataKind::LLVMDIBasicTypeMetadataKind
             | LLVMMetadataKind::LLVMDISubroutineTypeMetadataKind
             | LLVMMetadataKind::LLVMDIFileMetadataKind
-            | LLVMMetadataKind::LLVMDICompileUnitMetadataKind
             | LLVMMetadataKind::LLVMDILexicalBlockMetadataKind
             | LLVMMetadataKind::LLVMDILexicalBlockFileMetadataKind
             | LLVMMetadataKind::LLVMDINamespaceMetadataKind
--- a/tests/btf/assembly/c_void_enum.rs
+++ b/tests/btf/assembly/c_void_enum.rs
@@ -1,0 +1,19 @@
+// assembly-output: bpf-linker
+// no-prefer-dynamic
+// compile-flags: --crate-type bin -C link-arg=--emit=obj -C link-arg=--btf -C debuginfo=2
+
+#![no_std]
+#![no_main]
+
+use core::ffi::c_void;
+
+#[no_mangle]
+static mut FOO: *mut c_void = core::ptr::null_mut();
+
+#[panic_handler]
+fn panic(_info: &core::panic::PanicInfo) -> ! {
+    loop {}
+}
+
+// We check the BTF dump out of btfdump
+// CHECK-NOT: <ENUM> 'c_void'


tamird (changes requested):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/260)* status: 1 of 3 files reviewed, 5 unresolved discussions

___
*[`src/llvm/di.rs` line 76 at r2](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK2fxSDCs5Bz2MJ4p7y:-OK2fxSDCs5Bz2MJ4p7z:bdfz02c) ([raw file](https://github.com/aya-rs/bpf-linker/blob/d99bbd7ebac0e7e4c8b7e674a7253d321b300fe5/src/llvm/di.rs#L76)):*
> ```XML
> 
>     fn visit_mdnode_item(&mut self, item: Item) {
>         // guardrail preventing to visit non mdnode items
> ```

this kind of sucks. can `item.to_mdnode` return an option instead?
___
*[`src/llvm/di.rs` line 82 at r2](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK2gEVaCGYdNo-qE7Dd:-OK2aHjH9-LFhkZ-q7Mp:b-aycbbb) ([raw file](https://github.com/aya-rs/bpf-linker/blob/d99bbd7ebac0e7e4c8b7e674a7253d321b300fe5/src/llvm/di.rs#L82)):*
> ```XML
> 
>         let mdnode = item.as_mdnode();
>         match mdnode.clone().try_into().expect("MDNode is not Metadata") {
> ```

what's with this clone here?
___
*[`src/llvm/di.rs` line 90 at r2](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK2gAWH7Yjef15One3e:-OK2gAWH7Yjef15One3f:b-3uf2f6) ([raw file](https://github.com/aya-rs/bpf-linker/blob/d99bbd7ebac0e7e4c8b7e674a7253d321b300fe5/src/llvm/di.rs#L90)):*
> ```XML
>                         let name = di_composite_type
>                             .name()
>                             .map(|name| name.to_string_lossy().to_string());
> ```

i don't think conversion to lossy is appropriate here
___
*[`src/llvm/di.rs` line 104 at r2](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK2gJYi2wclXB7Xltlo:-OK2gJYi2wclXB7Xltlp:b-ggej6d) ([raw file](https://github.com/aya-rs/bpf-linker/blob/d99bbd7ebac0e7e4c8b7e674a7253d321b300fe5/src/llvm/di.rs#L104)):*
> ```XML
>                                             new.len(),
>                                             8,
>                                             // DWARF void
> ```

is this DW_ATE_void? write that, please.
___
*[`src/llvm/di.rs` line 113 at r2](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK2gPSV5BDdrchk5cyV:-OK2gPSV5BDdrchk5cyW:b-1d8cso) ([raw file](https://github.com/aya-rs/bpf-linker/blob/d99bbd7ebac0e7e4c8b7e674a7253d321b300fe5/src/llvm/di.rs#L113)):*
> ```XML
>                                     }
>                                 } else {
>                                     warn!("failed at replacing c_void enum, it might result in BTF parsing errors in kernels < 5.4")
> ```

this is a weird thing to do. on 32 bit size_in_bits won't be 8, but also this would all be unnecessary (right?). why log a warning then?

qjerome:
> > Reviewed 1 of 1 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/260)_ status: 1 of 3 files reviewed, 5 unresolved discussions
> 
> _[`src/llvm/di.rs` line 76 at r2](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK2fxSDCs5Bz2MJ4p7y:-OK2fxSDCs5Bz2MJ4p7z:bdfz02c) ([raw file](https://github.com/aya-rs/bpf-linker/blob/d99bbd7ebac0e7e4c8b7e674a7253d321b300fe5/src/llvm/di.rs#L76)):_
> 
> > ```
> >     fn visit_mdnode_item(&mut self, item: Item) {
> >         // guardrail preventing to visit non mdnode items
> > ```
> 
> this kind of sucks. can `item.to_mdnode` return an option instead?
Definitely possible ! Will fix accordingly.
> 
> _[`src/llvm/di.rs` line 82 at r2](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK2gEVaCGYdNo-qE7Dd:-OK2aHjH9-LFhkZ-q7Mp:b-aycbbb) ([raw file](https://github.com/aya-rs/bpf-linker/blob/d99bbd7ebac0e7e4c8b7e674a7253d321b300fe5/src/llvm/di.rs#L82)):_
> 
> > ```
> >         let mdnode = item.as_mdnode();
> >         match mdnode.clone().try_into().expect("MDNode is not Metadata") {
> > ```
> 
> what's with this clone here?
Will remove it's a leftover artifacts of test implementation.
> 
> _[`src/llvm/di.rs` line 90 at r2](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK2gAWH7Yjef15One3e:-OK2gAWH7Yjef15One3f:b-3uf2f6) ([raw file](https://github.com/aya-rs/bpf-linker/blob/d99bbd7ebac0e7e4c8b7e674a7253d321b300fe5/src/llvm/di.rs#L90)):_
> 
> > ```
> >                         let name = di_composite_type
> >                             .name()
> >                             .map(|name| name.to_string_lossy().to_string());
> > ```
> 
> i don't think conversion to lossy is appropriate here
Will fix that
> 
> _[`src/llvm/di.rs` line 104 at r2](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK2gJYi2wclXB7Xltlo:-OK2gJYi2wclXB7Xltlp:b-ggej6d) ([raw file](https://github.com/aya-rs/bpf-linker/blob/d99bbd7ebac0e7e4c8b7e674a7253d321b300fe5/src/llvm/di.rs#L104)):_
> 
> > ```
> >                                             new.len(),
> >                                             8,
> >                                             // DWARF void
> > ```
> 
> is this DW_ATE_void? write that, please. Will fix this.
> 
> _[`src/llvm/di.rs` line 113 at r2](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK2gPSV5BDdrchk5cyV:-OK2gPSV5BDdrchk5cyW:b-1d8cso) ([raw file](https://github.com/aya-rs/bpf-linker/blob/d99bbd7ebac0e7e4c8b7e674a7253d321b300fe5/src/llvm/di.rs#L113)):_
> 
> > ```
> >                                     }
> >                                 } else {
> >                                     warn!("failed at replacing c_void enum, it might result in BTF parsing errors in kernels < 5.4")
> > ```
> 
> this is a weird thing to do. on 32 bit size_in_bits won't be 8, but also this would all be unnecessary (right?). why log a warning then?

`c_void` Rust enum is `#repr[u8]` so its size should always be 8 bits. NB: the warning isn't on the `c_void` enum detection it is when we cannot fix the enum because we don't have the parent in which we can replace the `MDNode`. This was just to show somewhere that we couldn't fix the `c_void` enum. I neither wanted to left that else case unhandled nor panic on such case. So I thought emitting a warning was a good trade-off.

tamird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/260)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @qjerome)

___
*[`src/llvm/di.rs` line 104 at r2](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK2gJYi2wclXB7Xltlo:-OK69CZT9ljuCyIgKNXg:bhhdgdd) ([raw file](https://github.com/aya-rs/bpf-linker/blob/5abfe82aaa1d3a2f0121561b82b14271b9bb8225/src/llvm/di.rs#L104)):*
<details><summary><i>Previously, qjerome (Quentin JEROME) wrote…</i></summary><blockquote>

Will fix this.
</blockquote></details>

Actually, I no longer understand this comment. What is it anchored on?
___
*[`src/llvm/di.rs` line 113 at r2](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK2gPSV5BDdrchk5cyV:-OK68oMU-9AwH4-YGlJO:b-ffvvp8) ([raw file](https://github.com/aya-rs/bpf-linker/blob/5abfe82aaa1d3a2f0121561b82b14271b9bb8225/src/llvm/di.rs#L113)):*
> `c_void` Rust enum is `#repr[u8]` so its size should always be 8 bits.

on a 32 bit target?

anyway, I see now that this warning is logged when the item is not an operand. Please include that detail here.
___
*[`src/llvm/di.rs` line 90 at r6](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK69LbG349vIMbczjct:-OK69LbG349vIMbczjcu:bjx2szi) ([raw file](https://github.com/aya-rs/bpf-linker/blob/38c41e7cbf7730973b24d39a59d95d919ec7a159/src/llvm/di.rs#L90)):*
> ```XML
>                             if name == c"c_void" && di_composite_type.size_in_bits() == 8 {
>                                 if let Item::Operand(mut operand) = item {
>                                     let new = "void";
> ```

is there a reason we need to rename it? if yes, write a comment.

also, you seem to be creating this type every time you hit an operand of this type, rather than just once.
___
*[`src/llvm/di.rs` line 285 at r6](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK69_nw1ilyU-LdLddK:-OK69_nw1ilyU-LdLddL:b2u6wzy) ([raw file](https://github.com/aya-rs/bpf-linker/blob/38c41e7cbf7730973b24d39a59d95d919ec7a159/src/llvm/di.rs#L285)):*
> ```XML
> 
>         // if our item is also a MDNode
>         if item.is_mdnode() {
> ```

remove this check, it is in the method (and remove `is_mdnode`.

tamird (changes requested):
@qjerome please respond to every comment before requesting additional reviews.

> Reviewed 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-:-OK7DU_nB-dn8KbaM6p6:bqimews)* status: 2 of 3 files reviewed, 7 unresolved discussions (waiting on @qjerome)

___
*[`src/llvm/di.rs` line 118 at r7](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK7CZYV95zcLC0PcqvU:-OK7CZYV95zcLC0PcqvV:bu3w1wj) ([raw file](https://github.com/aya-rs/bpf-linker/blob/8f75409a03c0618a0f57e1514681766a2e9c94f1/src/llvm/di.rs#L118)):*
> ```XML
>     }
> 
>     /// Get or cache a LLVM DIBasicType given a [DIBasicType]
> ```

This comment should say something about returning a metadata ref for the given basic type.

Code-like references should be fenced e.g.

```
[`DIBasicType`]
```

Comments should be proper sentences. Please add a period.
___
*[`src/llvm/di.rs` line 123 at r7](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK7CW0Y3GOt04ipfHDm:-OK7CW0Y3GOt04ipfHDn:b-wu6ajg) ([raw file](https://github.com/aya-rs/bpf-linker/blob/8f75409a03c0618a0f57e1514681766a2e9c94f1/src/llvm/di.rs#L123)):*
> ```XML
>             .basic_types
>             .entry(di_bt)
>             .or_insert(di_bt.llvm_create_basic_type(self.builder))
> ```

this should be `or_insert_with`. The way you've written it `llvm_Create_basic_type` is *always* called.
___
*[`src/llvm/di.rs` line 139 at r7](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK7DPkP8IitGI0i7l3J:-OK7DPkP8IitGI0i7l3K:b-y6wdhp) ([raw file](https://github.com/aya-rs/bpf-linker/blob/8f75409a03c0618a0f57e1514681766a2e9c94f1/src/llvm/di.rs#L139)):*
> ```XML
>                         if let Some(name) = di_composite_type.name() {
>                             // we found the c_void enum
>                             if name == c"c_void" && di_composite_type.size_in_bits() == 8 {
> ```

shouldn't all this be using DIBasicType::Void instead of repeating all the details?
___
*[`src/llvm/di.rs` line 144 at r7](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK7D2eq9RuDeG3NQ2qx:-OK7D2eq9RuDeG3NQ2qy:b-ttqqhr) ([raw file](https://github.com/aya-rs/bpf-linker/blob/8f75409a03c0618a0f57e1514681766a2e9c94f1/src/llvm/di.rs#L144)):*
> ```XML
>                                     // get LLVM void DIBasicType
>                                     let void_bt = self.di_basic_type(DIBasicType::Void);
>                                     let _ =
> ```

what is this `_`?
___
*[`src/llvm/di.rs` line 310 at r7](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK7DJgo20t1kg42c0Bk:-OK7DJgo20t1kg42c0Bl:b-k0vxmq) ([raw file](https://github.com/aya-rs/bpf-linker/blob/8f75409a03c0618a0f57e1514681766a2e9c94f1/src/llvm/di.rs#L310)):*
> ```XML
>         };
> 
>         let first_visit = self.visited_nodes.insert(value_id);
> ```

is there a reason this moved?

qjerome:
> @qjerome please respond to every comment before requesting additional reviews.
> 
> > Reviewed 1 of 2 files at r2, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-:-OK7DU_nB-dn8KbaM6p6:bqimews)_ status: 2 of 3 files reviewed, 7 unresolved discussions (waiting on @qjerome)
> 
> _[`src/llvm/di.rs` line 118 at r7](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK7CZYV95zcLC0PcqvU:-OK7CZYV95zcLC0PcqvV:bu3w1wj) ([raw file](https://github.com/aya-rs/bpf-linker/blob/8f75409a03c0618a0f57e1514681766a2e9c94f1/src/llvm/di.rs#L118)):_
> 
> > ```
> >     }
> > 
> >     /// Get or cache a LLVM DIBasicType given a [DIBasicType]
> > ```
> 
> This comment should say something about returning a metadata ref for the given basic type.
> 
> Code-like references should be fenced e.g.
> 
> ```
> [`DIBasicType`]
> ```
> 
> Comments should be proper sentences. Please add a period.

Done

> 
> _[`src/llvm/di.rs` line 123 at r7](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK7CW0Y3GOt04ipfHDm:-OK7CW0Y3GOt04ipfHDn:b-wu6ajg) ([raw file](https://github.com/aya-rs/bpf-linker/blob/8f75409a03c0618a0f57e1514681766a2e9c94f1/src/llvm/di.rs#L123)):_
> 
> > ```
> >             .basic_types
> >             .entry(di_bt)
> >             .or_insert(di_bt.llvm_create_basic_type(self.builder))
> > ```
> 
> this should be `or_insert_with`. The way you've written it `llvm_Create_basic_type` is _always_ called.

Forgot about that HashMap trickery. Will change it ...

> 
> _[`src/llvm/di.rs` line 139 at r7](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK7DPkP8IitGI0i7l3J:-OK7DPkP8IitGI0i7l3K:b-y6wdhp) ([raw file](https://github.com/aya-rs/bpf-linker/blob/8f75409a03c0618a0f57e1514681766a2e9c94f1/src/llvm/di.rs#L139)):_
> 
> > ```
> >                         if let Some(name) = di_composite_type.name() {
> >                             // we found the c_void enum
> >                             if name == c"c_void" && di_composite_type.size_in_bits() == 8 {
> > ```
> 
> shouldn't all this be using DIBasicType::Void instead of repeating all the details?

No, `c_void` enum is the type in Rust code. `DIBasicType::Void` is what is used to create the void LLVM DIBasicType.

> 
> _[`src/llvm/di.rs` line 144 at r7](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK7D2eq9RuDeG3NQ2qx:-OK7D2eq9RuDeG3NQ2qy:b-ttqqhr) ([raw file](https://github.com/aya-rs/bpf-linker/blob/8f75409a03c0618a0f57e1514681766a2e9c94f1/src/llvm/di.rs#L144)):_
> 
> > ```
> >                                     // get LLVM void DIBasicType
> >                                     let void_bt = self.di_basic_type(DIBasicType::Void);
> >                                     let _ =
> > ```
> 
> what is this `_`?

This is because we don't use the return value of the function call. If you speak about while there's a line break ... IDK, ask rust-analyzer :)

> 
> _[`src/llvm/di.rs` line 310 at r7](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK7DJgo20t1kg42c0Bk:-OK7DJgo20t1kg42c0Bl:b-k0vxmq) ([raw file](https://github.com/aya-rs/bpf-linker/blob/8f75409a03c0618a0f57e1514681766a2e9c94f1/src/llvm/di.rs#L310)):_
> 
> > ```
> >         };
> > 
> >         let first_visit = self.visited_nodes.insert(value_id);
> > ```
> 
> is there a reason this moved?

Yes, instead of making the replacement in `visit_mdnode_item`, I opted for updating the `DISanitizer` `replace_operands` HashMap, because I find it a lot cleaner and this re-uses existing code. The only issue was the order the functions were called in `visit_item`, because `visit_mdnode_item` was called after the replacement took place our changes were not taken into account. So I had to change the order so that the replacement we make in `visit_mdnode_item` is taken into account.

tamird (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/260)* status: 2 of 3 files reviewed, 6 unresolved discussions (waiting on @qjerome)

___
*[`src/llvm/di.rs` line 118 at r7](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK7CZYV95zcLC0PcqvU:-OKBINef6nCeI7VZFxL7:btc330y) ([raw file](https://github.com/aya-rs/bpf-linker/blob/8f75409a03c0618a0f57e1514681766a2e9c94f1/src/llvm/di.rs#L118)):*
<details><summary><i>Previously, qjerome (Quentin JEROME) wrote…</i></summary><blockquote>

Done
</blockquote></details>

The same is true of DIBasicType.
___
*[`src/llvm/di.rs` line 144 at r7](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK7D2eq9RuDeG3NQ2qx:-OKBIZrT6fcdZ96aopbt:b-896zxi) ([raw file](https://github.com/aya-rs/bpf-linker/blob/8f75409a03c0618a0f57e1514681766a2e9c94f1/src/llvm/di.rs#L144)):*
<details><summary><i>Previously, qjerome (Quentin JEROME) wrote…</i></summary><blockquote>

This is because we don't use the return value of the function call. If you speak about while there's a line break ... IDK, ask rust-analyzer :)
</blockquote></details>

`let _` is discarding information. At the very least we should ascribe the type so I can tell if it's ok to throw this away or not.

qjerome:
> > _[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/260)_ status: 2 of 3 files reviewed, 6 unresolved discussions (waiting on @qjerome)
> 
> _[`src/llvm/di.rs` line 118 at r7](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK7CZYV95zcLC0PcqvU:-OKBINef6nCeI7VZFxL7:btc330y) ([raw file](https://github.com/aya-rs/bpf-linker/blob/8f75409a03c0618a0f57e1514681766a2e9c94f1/src/llvm/di.rs#L118)):_
> Previously, qjerome (Quentin JEROME) wrote…
> 
> > Done
> 
> The same is true of DIBasicType.

done

> 
> _[`src/llvm/di.rs` line 144 at r7](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK7D2eq9RuDeG3NQ2qx:-OKBIZrT6fcdZ96aopbt:b-896zxi) ([raw file](https://github.com/aya-rs/bpf-linker/blob/8f75409a03c0618a0f57e1514681766a2e9c94f1/src/llvm/di.rs#L144)):_
> Previously, qjerome (Quentin JEROME) wrote…
> 
> `let _` is discarding information. At the very least we should ascribe the type so I can tell if it's ok to throw this away or not.

This is now fixed because I had to change a bit that part of the code.

tamird (commented):
> Reviewed 2 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/260)* status: 3 of 4 files reviewed, 14 unresolved discussions (waiting on @qjerome)

___
*[`src/llvm/di.rs` line 118 at r7](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK7CZYV95zcLC0PcqvU:-OKRWpqB72mEZ9V8qZPs:bfhqml6) ([raw file](https://github.com/aya-rs/bpf-linker/blob/8f75409a03c0618a0f57e1514681766a2e9c94f1/src/llvm/di.rs#L118)):*
<details><summary><i>Previously, qjerome (Quentin JEROME) wrote…</i></summary><blockquote>

done
</blockquote></details>

Not done.
___
*[`src/llvm/types/di.rs` line 454 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRVk909CeJnuZ1Gv9o:-OKRVk909CeJnuZ1Gv9p:bj2qpu1) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/types/di.rs#L454)):*
> ```XML
> 
> impl DICompileUnit<'_> {
>     /// Constructs a new [`DICompileUnit`] from the given `value`.
> ```

s/value/value_ref/
___
*[`src/llvm/types/di.rs` line 469 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRVuCL4KMG3OvTL2Gt:-OKRVuCL4KMG3OvTL2Gu:bw0rvcc) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/types/di.rs#L469)):*
> ```XML
>     }
> 
>     pub fn enum_types(&self) -> Vec<DICompositeType> {
> ```

do we need to return a vector, or can we return `impl Iterator<Item=DICompositeType> + '_`?
___
*[`src/llvm/types/di.rs` line 489 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRWF7C7Ne8FJmaPEcN:-OKRWF7C7Ne8FJmaPEcO:bnf4gsj) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/types/di.rs#L489)):*
> ```XML
>     }
> 
>     pub fn replace_enum_types(&mut self, builder: LLVMDIBuilderRef, rep: Vec<DICompositeType>) {
> ```

this can be `&[DICompositeType]`, no need to force a vector
___
*[`src/llvm/types/di.rs` line 490 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRWIb_DAAiq4GxlO0e:-OKRWIb_DAAiq4GxlO0f:bzgyr7) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/types/di.rs#L490)):*
> ```XML
> 
>     pub fn replace_enum_types(&mut self, builder: LLVMDIBuilderRef, rep: Vec<DICompositeType>) {
>         let mut rep: Vec<LLVMMetadataRef> = rep.into_iter().map(|dct| dct.metadata_ref).collect();
> ```

nit: `Vec<_>`
___
*[`src/llvm/types/di.rs` line 493 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRWOiK2x_VhrvgXwFM:-OKRWOiK2x_VhrvgXwFN:beoct45) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/types/di.rs#L493)):*
> ```XML
> 
>         unsafe {
>             // we create an array with our saved enums
> ```

useless comments, here and below
___
*[`src/llvm/types/ir.rs` line 115 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRVOER1asRjZI3EAWx:-OKRVOER1asRjZI3EAWy:b-f49h46) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/types/ir.rs#L115)):*
> ```XML
>     DIDerivedType(DIDerivedType<'ctx>),
>     DISubprogram(DISubprogram<'ctx>),
>     DICompileUnit(#[allow(dead_code)] DICompileUnit<'ctx>),
> ```

what's with the dead code here?
___
*[`src/llvm/di.rs` line 58 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRWmeu2gax3VJn16in:-OKRWmeu2gax3VJn16io:blt7vj0) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/di.rs#L58)):*
> ```XML
>     }
> 
>     // DWARF encoding https://github.com/bminor/glibc/blob/2fe5e2af0995a6e6ee2c761e55e7596a3220d07c/sysdeps/generic/dwarf2.h#L375
> ```

How about https://llvm.org/docs/LangRef.html#dibasictype as a reference instead?
___
*[`src/llvm/di.rs` line 126 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRWujt9FLj_oE4J9uU:-OKRWujt9FLj_oE4J9uV:b-okp7je) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/di.rs#L126)):*
> ```XML
>     }
> 
>     fn visit_mdnode_item(&mut self, item: Item) {
> ```

item.as_mdnode() takes &self, why does this need to be Item rather than &Item?
___
*[`src/llvm/di.rs` line 504 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRX2GB8LhmpcGrDJxf:-OKRX2GB8LhmpcGrDJxg:b-436178) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/di.rs#L504)):*
> ```XML
>     }
> 
>     // Removes `c_void` Rust enum from all DICompileUnit.
> ```

what's going on here? did you mean to remove the changes in visit_mdnode_item? it looks like the same logic is here and there.

qjerome:
> > Reviewed 2 of 3 files at r10, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/260)_ status: 3 of 4 files reviewed, 14 unresolved discussions (waiting on @qjerome)
> 
> _[`src/llvm/di.rs` line 118 at r7](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK7CZYV95zcLC0PcqvU:-OKRWpqB72mEZ9V8qZPs:bfhqml6) ([raw file](https://github.com/aya-rs/bpf-linker/blob/8f75409a03c0618a0f57e1514681766a2e9c94f1/src/llvm/di.rs#L118)):_
> Previously, qjerome (Quentin JEROME) wrote…
> 
> Not done.

I hope it is OK now, otherwise maybe just write the comment you would like to see.

> 
> _[`src/llvm/types/di.rs` line 454 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRVk909CeJnuZ1Gv9o:-OKRVk909CeJnuZ1Gv9p:bj2qpu1) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/types/di.rs#L454)):_
> 
> > ```
> > impl DICompileUnit<'_> {
> >     /// Constructs a new [`DICompileUnit`] from the given `value`.
> > ```
> 
> s/value/value_ref/

done

> 
> _[`src/llvm/types/di.rs` line 469 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRVuCL4KMG3OvTL2Gt:-OKRVuCL4KMG3OvTL2Gu:bw0rvcc) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/types/di.rs#L469)):_
> 
> > ```
> >     }
> > 
> >     pub fn enum_types(&self) -> Vec<DICompositeType> {
> > ```
> 
> do we need to return a vector, or can we return `impl Iterator<Item=DICompositeType> + '_`?

done

> 
> _[`src/llvm/types/di.rs` line 489 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRWF7C7Ne8FJmaPEcN:-OKRWF7C7Ne8FJmaPEcO:bnf4gsj) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/types/di.rs#L489)):_
> 
> > ```
> >     }
> > 
> >     pub fn replace_enum_types(&mut self, builder: LLVMDIBuilderRef, rep: Vec<DICompositeType>) {
> > ```
> 
> this can be `&[DICompositeType]`, no need to force a vector

done

> 
> _[`src/llvm/types/di.rs` line 490 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRWIb_DAAiq4GxlO0e:-OKRWIb_DAAiq4GxlO0f:bzgyr7) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/types/di.rs#L490)):_
> 
> > ```
> >     pub fn replace_enum_types(&mut self, builder: LLVMDIBuilderRef, rep: Vec<DICompositeType>) {
> >         let mut rep: Vec<LLVMMetadataRef> = rep.into_iter().map(|dct| dct.metadata_ref).collect();
> > ```
> 
> nit: `Vec<_>`

done

> 
> _[`src/llvm/types/di.rs` line 493 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRWOiK2x_VhrvgXwFM:-OKRWOiK2x_VhrvgXwFN:beoct45) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/types/di.rs#L493)):_
> 
> > ```
> >         unsafe {
> >             // we create an array with our saved enums
> > ```
> 
> useless comments, here and below

done

> 
> _[`src/llvm/types/ir.rs` line 115 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRVOER1asRjZI3EAWx:-OKRVOER1asRjZI3EAWy:b-f49h46) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/types/ir.rs#L115)):_
> 
> > ```
> >     DIDerivedType(DIDerivedType<'ctx>),
> >     DISubprogram(DISubprogram<'ctx>),
> >     DICompileUnit(#[allow(dead_code)] DICompileUnit<'ctx>),
> > ```
> 
> what's with the dead code here?

It is to have a consistent Metadata enum with the new DICompileUnit type.
I can remove if you don't like it.

> 
> _[`src/llvm/di.rs` line 58 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRWmeu2gax3VJn16in:-OKRWmeu2gax3VJn16io:blt7vj0) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/di.rs#L58)):_
> 
> > ```
> >     }
> > 
> >     // DWARF encoding https://github.com/bminor/glibc/blob/2fe5e2af0995a6e6ee2c761e55e7596a3220d07c/sysdeps/generic/dwarf2.h#L375
> > ```
> 
> How about https://llvm.org/docs/LangRef.html#dibasictype as a reference instead?

done

> 
> _[`src/llvm/di.rs` line 126 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRWujt9FLj_oE4J9uU:-OKRWujt9FLj_oE4J9uV:b-okp7je) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/di.rs#L126)):_
> 
> > ```
> >     }
> > 
> >     fn visit_mdnode_item(&mut self, item: Item) {
> > ```
> 
> item.as_mdnode() takes &self, why does this need to be Item rather than &Item?

It facilitate the `Operand::replace` call a bit below, which requires a mutable variable.

> 
> _[`src/llvm/di.rs` line 504 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRX2GB8LhmpcGrDJxf:-OKRX2GB8LhmpcGrDJxg:b-436178) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/di.rs#L504)):_
> 
> > ```
> >     }
> > 
> >     // Removes `c_void` Rust enum from all DICompileUnit.
> > ```
> 
> what's going on here? did you mean to remove the changes in visit_mdnode_item? it looks like the same logic is here and there.

In `visit_mdnode_item` we replace `c_void` enum with a `DIBasicType`. However replacing does not remove the enum from existing in the `DICompileUnit` but `c_void` type is also changed in `DICompileUnit`. That type change in the `DICompileUnit` triggers an assertion failure in test with debug assertion because at some points it iterates over `enumTypes` and it expects to find a `DICompositeType` (used for enums). 

I didn't implement the `DICompileUnit` `c_void` enum removal in `visit_mdnode_item` because we need to be sure the removal happens at the very last (after all replacements have been done). Execution order we cannot really control in DFS.

tamird (commented):
> Reviewed 1 of 2 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/260)* status: 3 of 4 files reviewed, 5 unresolved discussions (waiting on @qjerome)

___
*[`src/llvm/di.rs` line 113 at r2](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK2gPSV5BDdrchk5cyV:-OKSJD7VA6Kzc-PczJez:b-yas1zh) ([raw file](https://github.com/aya-rs/bpf-linker/blob/5abfe82aaa1d3a2f0121561b82b14271b9bb8225/src/llvm/di.rs#L113)):*

> `c_void` Rust enum is `#repr[u8]` so its size should always be 8 bits.

on a 32 bit target?

anyway, I see now that this warning is logged when the item is not an operand. Please include that detail here.
</blockquote></details>

please address this.
___
*[`src/llvm/di.rs` line 126 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRWujt9FLj_oE4J9uU:-OKSJ33K28OfCIOemLqz:b-b69u70) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/di.rs#L126)):*
<details><summary><i>Previously, qjerome (Quentin JEROME) wrote…</i></summary><blockquote>

It facilitate the `Operand::replace` call a bit below, which requires `mut`.
</blockquote></details>

then why not pass a mut ref? if we're cloning this thing, what are we actually mutating? the copy?
___
*[`src/llvm/types/di.rs` line 489 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRWF7C7Ne8FJmaPEcN:-OKSJJT76V9XZmOXiX_t:b-hndczq) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/types/di.rs#L489)):*
<details><summary><i>Previously, qjerome (Quentin JEROME) wrote…</i></summary><blockquote>

done
</blockquote></details>

not done.
___
*[`src/llvm/types/ir.rs` line 115 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRVOER1asRjZI3EAWx:-OKSJ_bC2m9M8_LT9VFY:b-hxtnab) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/types/ir.rs#L115)):*
<details><summary><i>Previously, qjerome (Quentin JEROME) wrote…</i></summary><blockquote>

It is to have a consistent Metadata enum with the new DICompileUnit type.
I can remove if you don't like it.

</blockquote></details>

well i guess you need something or else the lifetime has nowhere to go.

qjerome:
> > Reviewed 1 of 2 files at r12, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/260)_ status: 3 of 4 files reviewed, 5 unresolved discussions (waiting on @qjerome)
> 
> _[`src/llvm/di.rs` line 113 at r2](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OK2gPSV5BDdrchk5cyV:-OKSJD7VA6Kzc-PczJez:b-yas1zh) ([raw file](https://github.com/aya-rs/bpf-linker/blob/5abfe82aaa1d3a2f0121561b82b14271b9bb8225/src/llvm/di.rs#L113)):_
> Previously, tamird (Tamir Duberstein) wrote…
> 
> please address this.

Sorry, I really don't get what you mean here ! I don't see why that enum would be different on 32-bits.

> 
> _[`src/llvm/di.rs` line 126 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRWujt9FLj_oE4J9uU:-OKSJ33K28OfCIOemLqz:b-b69u70) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/di.rs#L126)):_
> Previously, qjerome (Quentin JEROME) wrote…
> 
> then why not pass a mut ref? if we're cloning this thing, what are we actually mutating? the copy?

Correct ! That is fixed.

> 
> _[`src/llvm/types/di.rs` line 489 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRWF7C7Ne8FJmaPEcN:-OKSJJT76V9XZmOXiX_t:b-hndczq) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/types/di.rs#L489)):_
> Previously, qjerome (Quentin JEROME) wrote…
> 
> not done.

Sorry, my bad !

> 
> _[`src/llvm/types/ir.rs` line 115 at r11](https://reviewable.io/reviews/aya-rs/bpf-linker/260#-OKRVOER1asRjZI3EAWx:-OKSJ_bC2m9M8_LT9VFY:b-hxtnab) ([raw file](https://github.com/aya-rs/bpf-linker/blob/7b3876e821632e1fe5122bbe956d1fd1c14229e5/src/llvm/types/ir.rs#L115)):_
> Previously, qjerome (Quentin JEROME) wrote…
> 
> well i guess you need something or else the lifetime has nowhere to go.

Yes indeed, but my proposal was more to remove all `Metadata::DICompileUnit` related code as I we only need the `DICompileUnit` type so far. However I found it a bit strange and a bit confusing not to create this variant.

src/llvm/di.rs
```diff
@@ -21,13 +23,55 @@ use crate::llvm::{iter::*, types::di::DISubprogram};
 // backward compatibility
 const MAX_KSYM_NAME_LEN: usize = 128;
 
+#[derive(Hash, PartialEq, Eq, Clone, Copy)]
+enum DIBasicType {
+    I8,
+}
+
+impl DIBasicType {
+    fn llvm_create_basic_type(&self, builder: LLVMDIBuilderRef) -> LLVMMetadataRef {
```
vadorovsky, line 32: ...instead of returning raw pointers.

I see two options.

**Option 1**: Keeping the wrapper and enum separate, like:

````rust
pub struct DIBasicType<'ctx> {
    pub(super) metadata_ref: LLVMMetadataRef,
    _marker: PhantomData<&'ctx ()>,
}

pub enum DIBasicTypeKind {
    I8,
}
```

**Option 2**: Carry the wrapper in variants:

```
// DIBasicType like above

puib enum DIBasicTypeKind<'ctx> {
    I8(DIBasicType<'ctx>)
}
```

src/llvm/types/di.rs
```diff
@@ -430,3 +435,64 @@ impl DISubprogram<'_> {
...
+/// Represents the operands for a [`DICompileUnit`]. The enum values correspond
+/// to the operand indices within metadata nodes.
+#[repr(u32)]
+enum DICompileUnitOperand {
+    EnumTypes = 4,
+}
+
+/// Represents the debug information for a compile unit in LLVM IR.
+#[derive(Clone)]
+pub struct DICompileUnit<'ctx> {
```
vadorovsky, line 448: Good job with the wrapper here :+1:

vadorovsky (changes requested):

src/llvm/di.rs
```diff
@@ -21,13 +23,55 @@ use crate::llvm::{iter::*, types::di::DISubprogram};
 // backward compatibility
 const MAX_KSYM_NAME_LEN: usize = 128;
 
+#[derive(Hash, PartialEq, Eq, Clone, Copy)]
+enum DIBasicType {
```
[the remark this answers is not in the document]
qjerome: done

src/llvm/di.rs
```diff
@@ -21,13 +23,55 @@ use crate::llvm::{iter::*, types::di::DISubprogram};
...
+    I8,
```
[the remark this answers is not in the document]
qjerome: done

src/llvm/di.rs
qjerome: done

qjerome:
Hey @tamird @vadorovsky,

Maybe you forgot about this PR ! Is there anything blocking ?

Cheers,

src/llvm/di.rs
```diff
@@ -404,6 +438,58 @@ impl DISanitizer {
...
+        di_cus
+            .into_iter()
+            .map(|v| unsafe { DICompileUnit::from_value_ref(v) })
+            .collect()
+    }
+
+    // Removes `c_void` Rust enum from all DICompileUnit.
+    // After replacing `c_void` enum with a DIBasicType
+    // the DICompileUnit still references the `c_void` enum which
+    // triggers a casting assertion in LLVM.
```
vadorovsky, line 469: Please make it a proper docstring and add links to types:

```suggestion
    /// Removes `c_void` Rust enum from all [`DICompileUnit`].
    /// After replacing `c_void` enum with a [`DIBasicType`]
    /// the [`DICompileUnit`] still references the `c_void` enum which
    /// triggers a casting assertion in LLVM.
```

Also, the lines are a bit short, ~60 characters. Feel free to use up to ~80 per line.

src/llvm/di.rs
```diff
@@ -404,6 +438,58 @@ impl DISanitizer {
...
+    fn fix_di_compile_units(&mut self) {
+        for mut di_cu in self.di_compile_units() {
+            let tmp_cu = di_cu.clone();
+            let enum_types: Vec<_> = tmp_cu.enum_types().collect();
```
vadorovsky, line 473: You don't have to `collect` here. You are only using it as an iterator later, so you could skip collecting and just do

````rust
let new_enums = tmp_cu.enum_types()
    .filter(|e| {
         if let Some(name) = e.name() {
             // we filter c_void Rust enum
            if c"c_void" == name && e.size_in_bits() == 8 {
               return self.replaced_enums.contains(&e.value_id());
            }
        }
        false
   });:
```

You don't even have to `collect()` here :point_up_2: if you make `replace_enum_types` take an `impl IntoIterator<Item = DICompositeType>`, so you can pass both slices and iterators to that method. I will comment there as well.

If you manage to do it, you will end up with lazy iterators, which are consumed and looped over just once, instead of 3 loops like now.

src/llvm/types/di.rs
```diff
@@ -430,3 +444,136 @@ impl DISubprogram<'_> {
...
+            unsafe { LLVMGetNumOperands(llvm_enum_types) }
+        };
+
+        (0..llvm_enum_types_len).map(move |i| unsafe {
+            let enum_type = LLVMGetOperand(llvm_enum_types, i as u32);
+            DICompositeType::from_value_ref(enum_type)
+        })
+    }
+
+    pub fn replace_enum_types(&mut self, builder: LLVMDIBuilderRef, rep: &[DICompositeType]) {
```
vadorovsky, line 494:
```suggestion
    pub fn replace_enum_types(&mut self, builder: LLVMDIBuilderRef, rep: impl IntoIter<Item = DICompositeType>) {
```

or

```suggestion
    pub fn replace_enum_types<I>(&mut self, builder: LLVMDIBuilderRef, rep: I)
    where
        I: IntoIterator<Item = DICompositeType>,
    {
```

So you can pass both slices and lazy iterators here.

src/llvm/di.rs
qjerome: Done

src/llvm/di.rs
qjerome: There was a good reason to collect, it is because I used the number of enums as a trigger for enum replacement, not to replace enums when it is not needed. I will try to opt for another strategy but I don't know if we can implement that logic in a single loop ...

qjerome: I think we can do it in one loop if replace_enum_types takes a lazy iterator over MetadataRef but I think it is a bit dirty for an API.

src/llvm/types/di.rs
qjerome: Done

src/llvm/di.rs
qjerome: Reduced to 2 loops instead of 3 but I don't manage to make it in 1. Maybe I am lacking to much sleep

src/llvm/types/di.rs
```diff
@@ -430,3 +439,139 @@ impl DISubprogram<'_> {
...
+        match self {
+            Self::I8 => 8,
+        }
+    }
+
+    // DWARF encoding https://llvm.org/docs/LangRef.html#dibasictype
+    fn dwarf_type_encoding(&self) -> LLVMDWARFTypeEncoding {
+        match self {
+            // DW_ATE_signed
+            Self::I8 => 0x5,
```
vadorovsky, line 530: Use `gimli::DWE_ATE_signed`?

src/llvm/di.rs
```diff
@@ -438,6 +521,16 @@ impl Operand {
 }
 
 impl Item {
+    /// Returns the [`Item`] as [`MDNode`] only if [Item::is_mdnode] is `true` else `None`.
```
vadorovsky, line 524:
```suggestion
    /// Returns the [`Item`] as [`MDNode`] only if [`Item::is_mdnode`] is `true` else `None`.
```

src/llvm/di.rs
```diff
@@ -404,6 +434,59 @@ impl DISanitizer {
...
+    /// which triggers a casting assertion in LLVM.
+    fn fix_di_compile_units(&mut self) {
+        for mut di_cu in self.di_compile_units() {
+            let tmp_cu = di_cu.clone();
+            let enum_types = tmp_cu.enum_types();
+            let mut need_replace = false;
+            let new_enums: Vec<_> = enum_types
+                .into_iter()
+                .filter(|e| {
+                    let kind = unsafe { LLVMGetMetadataKind(e.metadata_ref) };
```
vadorovsky, line 474: Perhaps we should add a safe method `kind()` which returns `LLVMMetadataKind` and hides the `unsafe` call. Let's try to keep `unsafe` only in wrappers.

src/llvm/di.rs
```diff
@@ -404,6 +434,59 @@ impl DISanitizer {
...
+
+                    if matches!(kind, LLVMMetadataKind::LLVMDIBasicTypeMetadataKind) {
+                        need_replace = true;
+                        return false;
+                    }
+
+                    true
+                })
+                .collect();
```
vadorovsky, line 483: Since you're taking `IntoIterator` in `replace_enum_types`, I think this `collect` can go away?

qjerome: If we do that we will run `replace_enum_types` for every CU replacing each enums by itself which is a bit odd VS currently where we run it if and only if we know we have at least one enum to remove (which should be the case only when c_void enum is in CU).

src/llvm/di.rs
qjerome: fixed in https://github.com/aya-rs/bpf-linker/pull/260/commits/9a7e0c9ec03001523441c7ee96f3b03bd91593ba

src/llvm/types/di.rs
qjerome: fixed in https://github.com/aya-rs/bpf-linker/pull/260/commits/eb175e49f78395d4a56223cf74f30135b21847c8

src/llvm/di.rs
```diff
@@ -58,24 +61,51 @@ fn sanitize_type_name<T: AsRef<str>>(name: T) -> String {
...
             builder: unsafe { LLVMCreateDIBuilder(module) },
             visited_nodes: HashSet::new(),
             replace_operands: HashMap::new(),
             skipped_types: Vec::new(),
+            basic_types: HashMap::new(),
         }
     }
 
-    fn visit_mdnode(&mut self, mdnode: MDNode) {
+    /// Returns a metadata ref given a [`DIBasicType`].
```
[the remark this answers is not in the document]
qjerome: fixed in https://github.com/aya-rs/bpf-linker/pull/260/commits/c388fd9bace8d5abc994a807472c9ab15d290ab0

qjerome:
@vadorovsky @tamird @alessandrod : do you think this PR could be merge ?

src/llvm/di.rs
tamird: You're forcing a vector allocation here, and then you make another vector in `replace_enum_types`. Surely that is not necessary.

qjerome: I managed to remove one collect by using a peekable iterator. Please let me know if this solution is OK with you.

qjerome:
@tamird if you validate the last fix I'll squash for the final review.

src/llvm/di.rs
```diff
@@ -471,24 +471,16 @@ impl<'ctx> DISanitizer<'_> {
...
-                    if matches!(kind, LLVMMetadataKind::LLVMDIBasicTypeMetadataKind) {
-                        need_replace = true;
-                        return false;
-                    }
-
-                    true
-                })
-                .collect();
+            let mut enum_types = tmp_cu.enum_types().peekable();
+            let need_replace = enum_types.any(|ct| {
```
tamird: this is not how peekable works. you are passing a partially consumed iterator to replace_enum_types.

qjerome: Yes indeed ... Find the appropriate fix in the last commit.

qjerome:
o/ I don't know if you've seen it but I fixed the double vector allocation. Can you please validate or not this change so that I can proceed with squashing commits?

src/llvm/di.rs
tamird: Looks fine.

qjerome:
> I'm hesitant towards merging it, because:
> 
>     * I don't like the fact that it modifies `c_void` enums in two separate steps (first in `visit_mdnode` and then in `fix_di_compile_units`). That makes the code confusing - especially the part of having "enums" with `LLVMMetadataKind::LLVMDIBasicTypeMetadataKind` tag in between these two steps. I know it works this way for a reason - we first need to find an appropriate enum through DI, then remove it from global enums. But still, it adds complexity.
> 
>     * Given [the recent discussions about supporting binutils and lld as linkers for BPF targets in Rust](https://discord.com/channels/855676609003651072/1431944708468510800), I don't think we should introduce more fixups, unless absolutely necessary. I would like to eventually see 0 fixups in bpf-linker and upstream toolchains (GCC/binutils, LLVM/lld) being able to produce correct BTF.
> 
> 
> I appreciate your work on this, @qjerome. If you really care about <5.10 kernels, I think such a fixup could be implemented very easily in Aya, here:
> 
> https://github.com/aya-rs/aya/blob/948b8553ee72ab72d91c9a16e7e937a75eb0e155/aya-obj/src/btf/btf.rs#L490-L814
> 
> Doing that there would most likely nail down just to swapping different `Btf` enum variants. The downside of doing that there is that loading Rust eBPF programs with libbpf on kernel <=5.4 would not work, but I don't think there is anyone actually wanting to do that. 😛

OK, that was actually my first question when I started to work on this: whether I should fix that in Aya or in the `bpf-linker`. I've been told to work on the linker instead of fixing that in Aya ... During this work I didn't find any way to fix it in one shot, I am sorry to hear that this approach isn't good enough to be mergeable (as I invested some work into this). Anyway it was just an attempt at making the linker better support for BTF on older kernels.

vadorovsky:
> Sorry for letting this sit. Kernel 5.4 is quite old, and we don't have test coverage for it in aya. Can you help me understand that use case? I [recently](https://github.com/aya-rs/aya/pull/1251) added CI for 5.10 but I couldn't find an earlier LTS.

Unfortunately, Ubuntu supports even more ancient kernels than Debian does. The oldest Ubuntu LTS with legacy support in their [release cycle](https://ubuntu.com/about/release-cycle) is 14.04, which has kernel... 3.9. :skull: I personally hate to see this. But if someone is willing to do the work of supporting old kernels and manages to do so with a simple fixup on Aya and an integration test, I'm not going to oppose it. If it gets more complex, then my hesitation is going to grow.

> OK, that was actually my first question when I started to work on this: whether I should fix that in Aya or in the bpf-linker. I've been told to work on the linker instead of fixing that in Aya ...

I'm sorry that it turned out to be a lot of work for you and even more sorry if it was me suggesting that (I genuinely don't remember). At that moment we didn't know that:

* This fixup will become more complex than just swapping few operands, without interfering into the logic too much.
* We would consider supporting binutils and lld.

Given all of above, I don't think that linker should be the place that is aware of edge cases of old kernels. The least intrusive place is the loader (Aya) and that's where kernel-specific fixups already live.

tamird (changes requested):
Reviewing because the author requested a fresh maintainer review of this head.

The old-kernel failure is real, but this change puts a kernel-specific, two-stage LLVM debug-metadata rewrite in the linker for every BPF object. As https://github.com/aya-rs/bpf-linker/pull/260#issuecomment-3480796879 explains, kernel-specific BTF compatibility belongs in Aya’s existing BTF loader and sanitizer, not in toolchain-independent linker output. The new CHECK-NOT only verifies that c_void disappears; it does not establish that the affected kernel actually accepts the resulting BTF.

Please implement this in aya-obj’s existing fixup_and_sanitize path and cover the failing kernel with an integration test. If a demonstrated non-Aya consumer requires a linker fix instead, please establish that requirement before revisiting the ownership boundary.

All ten existing exact-head checks pass. This request concerns the design and missing behavioral coverage, not a CI failure.

— tamirdex

tamird:
Closing since this is likely the wrong approach.

PR #289, opened 2025-08-12
chore: Update to LLVM 21

tamird:
Latest Rust nightly switched to LLVM 21, so we need to update in order
to support rustc-llvm-proxy.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/289)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
           - beta
           - nightly
         llvm:
-          - 20
+          - 21
           - source
     name: rustc=${{ matrix.rust }} llvm=${{ matrix.llvm }}
     needs: llvm
--- a/.github/workflows/llvm.yml
+++ b/.github/workflows/llvm.yml
@@ -16,7 +16,7 @@ jobs:
       - id: ls-remote
         run: |
           set -euxo pipefail
-          value=$(git ls-remote https://github.com/aya-rs/llvm-project.git refs/heads/rustc/20.1-2025-02-13 | cut -f1)
+          value=$(git ls-remote https://github.com/aya-rs/llvm-project.git refs/heads/rustc/21.1-2025-08-01 | cut -f1)
           echo "sha=$value" >> "$GITHUB_OUTPUT"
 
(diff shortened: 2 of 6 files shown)


tamird:
> I filed a PR in llvm-sys https://gitlab.com/taricorp/llvm-sys.rs/-/merge_requests/55 If it gets merged quickly (1-2 days), perhaps we could include the bump in llvm-sys in new release.

Sounds good. Not a blocker for this PR, right?

banditopazzo:
in case of linking without the proxy, will be possible to link to a different llvm version?

vadorovsky:
> in case of linking without the proxy, will be possible to link to a different llvm version?

Linking to LLVM 20 will still likely work, even if we bump llvm-sys, just because there were not so many API changes and definitely none affecting us, so we are lucky. This PR manages to bump just LLVM without bumping the bindings.

With older than that, it would be tricky. I remember having problems using LLVM 19 when we switched to LLVM and llvm-sys 20.

Guaranteeing backwards compatibility would require a bit more work, out of the scope of this PR, but it doable. We would need to introduce some feature flags for picking different llvm-sys version like inkwell does:

https://github.com/TheDan64/inkwell/blob/master/Cargo.toml

I wouldn't go as far as them with supporting versions down to 8, but doing so 2 versions before (let's say, to 19) would make sense IMO.

Speaking of which - I think after doing so, we could figure out an experimental support of stable Rust version, something you've been asking for many times. :smile: In #282, I added an xtask command `build-std` for building the stdlib for BPF target, which is then permanently stored in a toolchain dir. After all, the "tier 3 support" of BPF is just something affecting the way Rust team releases their binaries. But there is nothing stopping people from building stable compilers with BPF targets or building BPF targets separately.

vadorovsky:
I managed to fix fixups, as mentioned here https://github.com/aya-rs/bpf-linker/issues/286#issuecomment-3188122319. CI should become green.

tamird:
Nice! Looks like generics is still broken though.

vadorovsky:
@tamird Green

PR #295, opened 2025-08-20
add in memory linking

banditopazzo:
I ended my first iteration for memory linking and api changes.

To recap my goals:
1. in memory linking
2. and since I am touching the external apis, add the possibility of reusing the linker instance avoiding a full initialization each time (in my use case I have to basically link in a for loop)

Major changes:
- better/safer wrappers for llvm objects (custom `Drop` in every wrapper, lifetimes, mutability)
- api change to have 2 main methods: link_to_file, link_to_buffer
- linker accepts both files and in-memory data at the same time
- moved some linker methods to standalone functions

Open points or missing:
- from what I understand the target_machine could be created early and stored in the linker instance, in theory it's based on the cli options
- names of the new structures
- `dump_module` should be moved to link methods

I am not surprised if I'm forgetting something. 

In any case this should be enough at least to have an idea of changes I am proposing

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/295)

### DIFF
--- a/src/bin/bpf-linker.rs
+++ b/src/bin/bpf-linker.rs
@@ -14,7 +14,7 @@ use std::{
     feature = "rust-llvm-21"
 ))]
 use aya_rustc_llvm_proxy as _;
-use bpf_linker::{Cpu, Linker, LinkerOptions, OptLevel, OutputType};
+use bpf_linker::{Cpu, Linker, LinkerInput, LinkerOptions, OptLevel, OutputType};
 use clap::{
     builder::{PathBufValueParser, TypedValueParser as _},
     error::ErrorKind,
@@ -271,15 +271,11 @@ fn main() -> anyhow::Result<()> {
 
     let export_symbols = export_symbols.map(fs::read_to_string).transpose()?;
 
-    // TODO: the data is owned by this call frame; we could make this zero-alloc.
     let export_symbols = export_symbols
         .as_deref()
         .into_iter()
         .flat_map(str::lines)
-        .map(str::to_owned)
-        .chain(export)
-        .map(Into::into)
-        .collect();
+        .chain(export.iter().map(String::as_str));
 
     let output_type = match *emit.as_slice() {
         [] => unreachable!("emit has a default value"),
@@ -294,11 +290,7 @@ fn main() -> anyhow::Result<()> {
         target,
         cpu,
         cpu_features,
-        inputs,
-        output,
-        output_type,
         optimize,
-        export_symbols,
         unroll_loops,
         ignore_inline_never,
         llvm_args,
@@ -312,7 +304,11 @@ fn main() -> anyhow::Result<()> {
         linker.set_dump_module_path(path);
     }
 
-    linker.link()?;
+    let inputs = inputs
+        .iter()
+        .map(|p| LinkerInput::new_from_file(p.as_path()));
+
+    linker.link_to_file(inputs, &output, output_type, export_symbols)?;
 
     if fatal_errors && linker.has_errors() {
         return Err(anyhow::anyhow!(
--- a/src/linker.rs
+++ b/src/linker.rs
@@ -3,7 +3,8 @@ use std::{
     collections::HashSet,
     ffi::{CStr, CString, OsStr},
     fs::File,
-    io::{self, Read, Seek as _},
+    io::{self, Read, Seek},
+    ops::Deref,
     os::unix::ffi::OsStrExt as _,
     path::{Path, PathBuf},
     str::{self, FromStr},
@@ -17,7 +18,7 @@ use llvm_sys::{
 use thiserror::Error;
 use tracing::{debug, error, info, warn};
 
-use crate::llvm::{self, LLVMContext, LLVMModule, LLVMTargetMachine};
+use crate::llvm::{self, LLVMContext, LLVMModule, LLVMTargetMachine, MemoryBuffer};
 
 /// Linker error
 #[derive(Debug, Error)]
@@ -141,6 +142,59 @@ pub enum OptLevel {
     SizeMin,
 }
 
+pub struct FileInput<'a> {
+    path: &'a Path,
+}
+
+pub struct BufferInput<'a> {
+    name: &'a str,
+    bytes: &'a [u8],
+}
+
+pub enum LinkerInput<'a> {
+    File(FileInput<'a>),
+    Buffer(BufferInput<'a>),
+}
+
+impl<'a> LinkerInput<'a> {
+    pub fn new_from_file(path: &'a Path) -> Self {
+        LinkerInput::File(FileInput { path })
+    }
+
+    pub fn new_from_buffer(name: &'a str, bytes: &'a [u8]) -> Self {
+        LinkerInput::Buffer(BufferInput { name, bytes })
+    }
+}
+
+enum InputReader<'a> {
+    File {
+        path: &'a Path,
+        file: File,
+    },
+    Buffer {
+        name: &'a str,
+        cursor: io::Cursor<&'a [u8]>,
+    },
+}
+
+impl Seek for InputReader<'_> {
+    fn seek(&mut self, pos: io::SeekFrom) -> io::Result<u64> {
+        match self {
+            InputReader::File { file, .. } => file.seek(pos),
+            InputReader::Buffer { cursor, .. } => cursor.seek(pos),
+        }
+    }
+}
+
+impl Read for InputReader<'_> {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        match self {
+            InputReader::File { file, .. } => file.read(buf),
+            InputReader::Buffer { cursor, .. } => cursor.read(buf),
+        }
+    }
+}
+
 /// Linker input type
 #[derive(Clone, Copy, Debug, PartialEq)]
 enum InputType {
@@ -192,16 +246,8 @@ pub struct LinkerOptions {
     pub cpu: Cpu,
     /// Cpu features.
     pub cpu_features: CString,
-    /// Input files. Can be bitcode, object files with embedded bitcode or archive files.
-    pub inputs: Vec<PathBuf>,
-    /// Where to save the output.
-    pub output: PathBuf,
-    /// The format to output.
-    pub output_type: OutputType,
     /// Optimization level.
     pub optimize: OptLevel,
-    /// Set of symbol names to export.
-    pub export_symbols: HashSet<Cow<'static, str>>,
     /// Whether to aggressively unroll loops. Useful for older kernels that don't support loops.
     pub unroll_loops: bool,
     /// Remove `noinline` attributes from functions. Useful for kernels before 5.8 that don't
@@ -254,8 +300,134 @@ impl Linker {
         self.dump_module = Some(path.as_ref().to_path_buf())
     }
 
+    /// Link and generate the output code to file.
+    ///
+    /// # Example
+    ///
+    /// ```rust,no_run
+    /// # use std::{collections::HashSet, path::Path, borrow::Cow, ffi::CString};
+    /// # use bpf_linker::{Cpu, Linker, LinkerInput, LinkerOptions, OptLevel, OutputType};
+    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// let path = Path::new("/path/to/object-or-bitcode");
+    /// let bytes: &[u8] = &[]; // An in memory object/bitcode
+    /// # let options = LinkerOptions {
+    /// #     target: None,
+    /// #     cpu: Cpu::Generic,
+    /// #     cpu_features: CString::default(),
+    /// #     optimize: OptLevel::Default,
+    /// #     unroll_loops: false,
+    /// #     ignore_inline_never: false,
+    /// #     llvm_args: vec![],
+    /// #     disable_expand_memcpy_in_order: false,
+    /// #     disable_memory_builtins: false,
+    /// #     allow_bpf_trap: false,
+    /// #     btf: false,
+    /// # };
+    /// # let linker = Linker::new(options);
+    ///
+    /// let export_symbols = ["my_sym_1", "my_sym_2"];
+    ///
+    /// linker.link_to_file(
+    ///     [
+    ///         LinkerInput::new_from_file(path),
+    ///         LinkerInput::new_from_buffer("my buffer", bytes), // In memory buffer needs a name
+    ///     ],
+    ///     "/path/to/output",
+    ///     OutputType::Object,
+    ///     export_symbols,
+    /// )?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn link_to_file<'i, 'a, I, P, E>(
+        &self,
+        inputs: I,
+        output: P,
+        output_type: OutputType,
+        export_symbols: E,
+    ) -> Result<(), LinkerError>
+    where
+        I: IntoIterator<Item = LinkerInput<'i>>,
+        E: IntoIterator<Item = &'a str>,
+        P: AsRef<Path>,
+    {
+        let (linked_module, target_machine) = self.link(inputs, export_symbols)?;
+        codegen_to_file(
+            &linked_module,
+            &target_machine,
+            output.as_ref(),
+            output_type,
+        )?;
+        Ok(())
+    }
+
+    /// Link and generate the output code to an in-memory buffer.
+    ///
+    /// # Example
+    ///
+    /// ```rust,no_run
+    /// # use std::{collections::HashSet, path::Path, borrow::Cow, ffi::CString};
+    /// # use bpf_linker::{Cpu, Linker, LinkerInput, LinkerOptions, OptLevel, OutputType};
+    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// let path = Path::new("/path/to/object-or-bitcode");
+    /// let bytes: &[u8] = &[]; // An in memory object/bitcode
+    /// # let options = LinkerOptions {
+    /// #     target: None,
+    /// #     cpu: Cpu::Generic,
+    /// #     cpu_features: CString::default(),
+    /// #     optimize: OptLevel::Default,
+    /// #     unroll_loops: false,
+    /// #     ignore_inline_never: false,
+    /// #     llvm_args: vec![],
+    /// #     disable_expand_memcpy_in_order: false,
+    /// #     disable_memory_builtins: false,
+    /// #     allow_bpf_trap: false,
+    /// #     btf: false,
+    /// # };
+    /// # let linker = Linker::new(options);
+    ///
+    /// let export_symbols = ["my_sym_1", "my_sym_2"];
+    ///
+    /// let out_buf = linker.link_to_buffer(
+    ///     [
+    ///         LinkerInput::new_from_file(path),
+    ///         LinkerInput::new_from_buffer("my buffer", bytes), // In memory buffer needs a name
+    ///     ],
+    ///     OutputType::Bitcode,
+    ///     export_symbols,
+    /// )?;
+    ///
+    /// // Use the buffer as slice of u8
+    /// let bytes = out_buf.as_slice();
+    /// println!("Linked {} bytes into memory)", bytes.len());
+    ///
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn link_to_buffer<'i, 'a, I, E>(
+        &self,
+        inputs: I,
+        output_type: OutputType,
+        export_symbols: E,
+    ) -> Result<LinkerOutput, LinkerError>
+    where
+        I: IntoIterator<Item = LinkerInput<'i>>,
+        E: IntoIterator<Item = &'a str>,
+    {
+        let (linked_module, target_machine) = self.link(inputs, export_symbols)?;
+        codegen_to_buffer(&linked_module, &target_machine, output_type)
+    }
+
     /// Link and generate the output code.
-    pub fn link(&mut self) -> Result<(), LinkerError> {
+    fn link<'ctx, 'i, 'a, I, E>(
+        &'ctx self,
+        inputs: I,
+        export_symbols: E,
+    ) -> Result<(LLVMModule<'ctx>, LLVMTargetMachine), LinkerError>
+    where
+        I: IntoIterator<Item = LinkerInput<'i>>,
+        E: IntoIterator<Item = &'a str>,
+    {
         let Self {
             options,
             context,
@@ -263,10 +435,34 @@ impl Linker {
             ..
         } = self;
 
-        let mut module = link_modules(context, &options.inputs)?;
+        let inputs = inputs
+            .into_iter()
+            .map(|value| match value {
+                LinkerInput::File(file_input) => {
+                    let FileInput { path } = file_input;
+
+                    let file = File::open(path)
+                        .map_err(|err| LinkerError::IoError(path.to_owned(), err))?;
+                    Ok(InputReader::File { path, file })
+                }
+                LinkerInput::Buffer(buffer_input) => {
+                    let BufferInput { name, bytes } = buffer_input;
+
+                    Ok(InputReader::Buffer {
+                        name,
+                        cursor: io::Cursor::new(bytes),
+                    })
+                }
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let mut module = link_modules(context, inputs)?;
+
         let target_machine = create_target_machine(options, &module)?;
-        if let Some(path) = &dump_module {
-            std::fs::create_dir_all(path).map_err(|err| LinkerError::IoError(path.clone(), err))?;
+
+        if let Some(path) = dump_module {
+            std::fs::create_dir_all(path)
+                .map_err(|err| LinkerError::IoError(path.to_owned(), err))?;
         }
         if let Some(path) = dump_module {
             // dump IR before optimization
@@ -281,7 +477,7 @@ impl Linker {
             context,
             &target_machine,
             &mut module,
-            &options.export_symbols,
+            export_symbols,
         )?;
         if let Some(path) = dump_module {
             // dump IR before optimization
@@ -291,13 +487,8 @@ impl Linker {
                 .write_ir_to_path(&path)
                 .map_err(LinkerError::WriteIRError)?;
         };
-        codegen(
-            &module,
-            &target_machine,
-            &options.output,
-            options.output_type,
-        )?;
-        Ok(())
+
+        Ok((module, target_machine))
     }
 
     pub fn has_errors(&self) -> bool {
@@ -305,24 +496,32 @@ impl Linker {
     }
 }
 
-fn link_modules<'ctx>(
+fn link_modules<'ctx, 'i, I>(
     context: &'ctx LLVMContext,
-    inputs: &[PathBuf],
-) -> Result<LLVMModule<'ctx>, LinkerError> {
+    inputs: I,
+) -> Result<LLVMModule<'ctx>, LinkerError>
+where
+    I: IntoIterator<Item = InputReader<'i>>,
+{
     let mut module = context
         .create_module(c"linked_module")
         .ok_or(LinkerError::CreateModuleError)?;
 
     // buffer used to perform file type detection
     let mut buf = [0u8; 8];
-    for path in inputs {
-        let mut file = File::open(path).map_err(|e| LinkerError::IoError(path.clone(), e))?;
+    for mut input in inputs {
+        let path = match input {
+            InputReader::File { path, .. } => path.into(),
+            InputReader::Buffer { name, .. } => PathBuf::from(format!("in_memory::{}", name)),
+        };
 
         // determine whether the input is bitcode, ELF with embedded bitcode, an archive file
         // or an invalid file
-        file.read_exact(&mut buf)
+        input
+            .read_exact(&mut buf)
             .map_err(|e| LinkerError::IoError(path.clone(), e))?;
-        file.rewind()
+        input
+            .rewind()
             .map_err(|e| LinkerError::IoError(path.clone(), e))?;
         let in_type =
             detect_input_type(&buf).ok_or_else(|| LinkerError::InvalidInputType(path.clone()))?;
@@ -332,7 +531,7 @@ fn link_modules<'ctx>(
                 info!("linking archive {:?}", path);
 
                 // Extract the archive and call link_reader() for each item.
-                let mut archive = Archive::new(file);
+                let mut archive = Archive::new(input);
                 while let Some(Ok(item)) = archive.next_entry() {
                     let name = PathBuf::from(OsStr::from_bytes(item.header().identifier()));
                     info!("linking archive item {:?}", name);
@@ -347,15 +546,13 @@ fn link_modules<'ctx>(
                             warn!("ignoring archive item {:?}: no embedded bitcode", name);
                             continue;
                         }
-                        Err(_) => {
-                            return Err(LinkerError::LinkArchiveModuleError(path.clone(), name))
-                        }
+                        Err(_) => return Err(LinkerError::LinkArchiveModuleError(path, name)),
                     };
                 }
             }
             ty => {
                 info!("linking file {:?} type {}", path, ty);
-                match link_reader(context, &mut module, path, file, Some(ty)) {
+                match link_reader(context, &mut module, &path, input, Some(ty)) {
                     Ok(_) => {}
                     Err(LinkerError::InvalidInputType(_)) => {
                         info!("ignoring file {:?}: invalid type", path);
@@ -468,13 +665,16 @@ fn create_target_machine(
     Ok(target_machine)
 }
 
-fn optimize<'ctx>(
+fn optimize<'ctx, 'a, E>(
     options: &LinkerOptions,
     context: &'ctx LLVMContext,
     target_machine: &LLVMTargetMachine,
     module: &mut LLVMModule<'ctx>,
-    export_symbols: &HashSet<Cow<'static, str>>,
-) -> Result<(), LinkerError> {
+    export_symbols: E,
+) -> Result<(), LinkerError>
+where
+    E: IntoIterator<Item = &'a str>,
+{
     let LinkerOptions {
         disable_memory_builtins,
         optimize,
@@ -483,8 +683,10 @@ fn optimize<'ctx>(
         ..
     } = options;
 
-    let mut export_symbols: HashSet<Cow<'_, [u8]>> =
-        export_symbols.iter().map(|s| s.as_bytes().into()).collect();
+    let mut export_symbols: HashSet<Cow<'_, [u8]>> = export_symbols
+        .into_iter()
+        .map(|s| Cow::Borrowed(s.as_bytes()))
+        .collect();
 
     if !disable_memory_builtins {
         export_symbols.extend(
@@ -521,7 +723,7 @@ fn optimize<'ctx>(
     Ok(())
 }
 
-fn codegen(
+fn codegen_to_file(
     module: &LLVMModule<'_>,
     target_machine: &LLVMTargetMachine,
     output: &Path,
@@ -543,6 +745,27 @@ fn codegen(
             .emit_to_file(module, &output, LLVMCodeGenFileType::LLVMObjectFile)
             .map_err(LinkerError::EmitCodeError),
     }
+}
+
+fn codegen_to_buffer(
+    module: &LLVMModule<'_>,
+    target_machine: &LLVMTargetMachine,
+    output_type: OutputType,
+) -> Result<LinkerOutput, LinkerError> {
+    let memory_buffer = match output_type {
+        OutputType::Bitcode => module.write_bitcode_to_memory(),
+        OutputType::LlvmAssembly => module.write_ir_to_memory(),
+        OutputType::Assembly => target_machine
+            .emit_to_memory_buffer(module, LLVMCodeGenFileType::LLVMAssemblyFile)
+            .map_err(LinkerError::EmitCodeError)?,
+        OutputType::Object => target_machine
+            .emit_to_memory_buffer(module, LLVMCodeGenFileType::LLVMObjectFile)
+            .map_err(LinkerError::EmitCodeError)?,
+    };
+
+    Ok(LinkerOutput {
+        inner: memory_buffer,
+    })
 }
 
 fn llvm_init(
@@ -660,5 +883,29 @@ fn detect_input_type(data: &[u8]) -> Option<InputType> {
                 None
             }
         }
+    }
+}
+
+pub struct LinkerOutput {
+    inner: MemoryBuffer,
+}
+
+impl LinkerOutput {
+    pub fn as_slice(&self) -> &[u8] {
+        self.inner.as_slice()
+    }
+}
+
+impl AsRef<[u8]> for LinkerOutput {
+    fn as_ref(&self) -> &[u8] {
+        self.as_slice()
+    }
+}
+
+impl Deref for LinkerOutput {
+    type Target = [u8];
+
+    fn deref(&self) -> &Self::Target {
+        self.as_slice()
     }
 }
--- a/src/llvm/mod.rs
+++ b/src/llvm/mod.rs
@@ -44,6 +44,7 @@ use llvm_sys::{
 use tracing::{debug, error};
 pub(crate) use types::{
     context::{InstalledDiagnosticHandler, LLVMContext},
+    memory_buffer::MemoryBuffer,
     module::LLVMModule,
     target_machine::LLVMTargetMachine,
 };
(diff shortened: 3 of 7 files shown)


banditopazzo:
another question is what to do with `libs` path list: to remove or to handle

src/llvm/context.rs
```diff
@@ -0,0 +1,41 @@
+use std::{ffi::CString, marker::PhantomData};
+
+use llvm_sys::{
+    core::{LLVMContextCreate, LLVMContextDispose, LLVMModuleCreateWithNameInContext},
+    prelude::LLVMContextRef,
+};
+
+use crate::llvm::LLVMModuleWrapped;
+
+pub struct LLVMContextWrapped {
```
vadorovsky: Awesome job with adding new wrappers! However, I think we should keep them in the `src/llvm/types` folder. And let's just name it `LLVMContext`.

src/llvm/memory_buffer.rs
```diff
@@ -0,0 +1,34 @@
+use core::slice;
+
+use llvm_sys::{
+    core::{LLVMDisposeMemoryBuffer, LLVMGetBufferSize, LLVMGetBufferStart},
+    prelude::LLVMMemoryBufferRef,
+};
+
+pub struct MemoryBufferWrapped {
```
vadorovsky: Same here - let's keep it in the `types` folder and drop the `Wrapped` suffix.

src/llvm/mod.rs
```diff
@@ -375,3 +346,14 @@ fn mdstring_to_str<'a>(mdstring: LLVMValueRef) -> &'a str {
     let ptr = unsafe { LLVMGetMDString(mdstring, &mut len) };
     unsafe { str::from_utf8(slice::from_raw_parts(ptr as *const c_uchar, len as usize)).unwrap() }
 }
+
+pub unsafe fn set_diagnostic_handler<T: LLVMDiagnosticHandler>(
```
vadorovsky: That's probably a good candidate for becoming a method of the `LLVMContext` wrapper

vadorovsky (commented):
Do you think you could split out the wrapper parts as a separate change? If so, we could merge it fast, and then review just the in-memory part of the code (which I still need to play with and review properly).

alessandrod:
> Hi @alessandrod , have you had a chance to take a look at this draft? I'd appreciate your thoughts on whether the changes work for you and if you like the direction it’s heading in.

gonna take a look today, sorry for the delay!

src/linker.rs
```diff
@@ -226,345 +268,84 @@ pub struct LinkerOptions {
...
-        }
+            context,
+            diagnostic_handler,
+        })
+    }
+
+    /// Link and generate the output code to file.
+    pub fn link_to_file(
+        &self,
+        inputs: Vec<LinkerInput>,
```
alessandrod: nit: this should probably be `&[LinkerInput]`

src/linker.rs
```diff
@@ -226,345 +268,84 @@ pub struct LinkerOptions {
...
     }
 
     /// Link and generate the output code.
-    pub fn link(&mut self) -> Result<(), LinkerError> {
-        self.llvm_init();
-        self.link_modules()?;
-        self.create_target_machine()?;
+    fn link<'ctx>(
+        &'ctx self,
+        inputs: Vec<LinkerInput>,
```
alessandrod: same here, slice?

src/linker.rs
```diff
@@ -633,3 +414,333 @@ fn detect_input_type(data: &[u8]) -> Option<InputType> {
...
+                target_machine.codegen_to_file(module, &output, LLVMCodeGenFileType::LLVMObjectFile)
+            }
+            .map_err(LinkerError::EmitCodeError)
+        }
+    }
+}
+
+fn link_modules<'ctx>(
+    context: &'ctx LLVMContextWrapped,
+    inputs: Vec<LinkerInput>,
```
alessandrod: slice?

src/linker.rs
```diff
@@ -226,345 +268,84 @@ pub struct LinkerOptions {
...
+        export_symbols: HashSet<Cow<'static, str>>,
+    ) -> Result<(), LinkerError> {
+        let (linked_module, target_machine) = self.link(inputs, &export_symbols)?;
+        codegen_to_file(&linked_module, &target_machine, &output, output_type)?;
+        Ok(())
+    }
+
+    pub fn link_to_buffer(
+        &self,
+        inputs: Vec<LinkerInput>,
```
alessandrod: slice?

src/linker.rs
```diff
@@ -226,345 +268,84 @@ pub struct LinkerOptions {
...
+        })
+    }
+
+    /// Link and generate the output code to file.
+    pub fn link_to_file(
+        &self,
+        inputs: Vec<LinkerInput>,
+        output: &Path,
+        output_type: OutputType,
+        export_symbols: HashSet<Cow<'static, str>>,
```
alessandrod: take by ref?

src/linker.rs
```diff
@@ -633,3 +414,333 @@ fn detect_input_type(data: &[u8]) -> Option<InputType> {
...
+        inner: memory_buffer,
+    })
+}
+
+// To avoid leaking of internal LLVM wrapper
+pub struct LinkerOutput {
+    inner: MemoryBufferWrapped,
+}
+
+impl LinkerOutput {
```
alessandrod, line 893: should probably also implement `AsRef<[u8]>`

src/llvm/module.rs
```diff
@@ -0,0 +1,74 @@
...
+
+use crate::llvm::{MemoryBufferWrapped, Message};
+
+pub struct LLVMModuleWrapped<'ctx> {
+    pub(super) module: LLVMModuleRef,
+    pub(super) _marker: PhantomData<&'ctx super::LLVMContextWrapped>,
+}
+
+impl<'ctx> LLVMModuleWrapped<'ctx> {
+    pub unsafe fn get_target(&self) -> *const i8 {
```
alessandrod: should be as `*const c_char`

examples/basic.rs
```diff
@@ -0,0 +1,54 @@
+use std::{collections::HashSet, path::Path};
+
+use bpf_linker::{Cpu, Linker, LinkerInput, LinkerOptions, OptLevel, OutputType};
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let path = Path::new("/path/to/object-or-bitcode");
```
alessandrod: wouldn't this example be better as Examples in the rustdocs for the respective methods?

banditopazzo:
I rebased on the main and squashed to remove useless commits.

The I addressed most of the review comments.

@vadorovsky: I tried to create a new PR without the in memory linking but it ended just like this, the in memory linking is marginal. I think it doesn't make the review process faster, but if really want as separate PR, I will do it.

@alessandrod: taking a slice instead of the `Vec<LinkerInput>` as input wouldn't work in my opinion. The `LinkerInput` objects have a mutable state inside because of the cursor they use for the `Read`, so it needs to be a `&mut [LinkerInput]`. I don't see any advantage doing this, considering `LinkerInput` is just a type that we use to accept a mixed vector of both files and in memory data

src/bin/bpf-linker.rs
```diff
@@ -118,10 +118,6 @@ struct CommandLine {
     #[clap(long)]
     allow_bpf_trap: bool,
 
-    /// Add a directory to the library search path
```
alessandrod, line 120: why this change? bpf-linker needs to stay compatible with rustc, we can't change
the CLI

src/linker.rs
```diff
@@ -229,354 +266,169 @@ pub struct LinkerOptions {
...
+    /// #     disable_expand_memcpy_in_order: false,
+    /// #     disable_memory_builtins: false,
+    /// #     allow_bpf_trap: false,
+    /// #     btf: false,
+    /// # };
+    /// # let linker = Linker::new(options)?;
+    ///
+    /// // Link to a file
+    /// linker.link_to_file(
+    ///     vec![
```
alessandrod: this is why Vec is bad: you're forcing people to allocate here, while

```
[
	LinkerInput::try_from(path),
]
```

should work too

something like ```impl IntoIterator<Item = impl BorrowMut<LinkerInput<'a>>>```
should work

src/linker.rs
```diff
@@ -141,6 +137,58 @@ pub enum OptLevel {
     SizeMin,
 }
 
+pub struct FileInput<'a> {
+    path: &'a Path,
+    file: File,
+}
+
+pub struct BytesInput<'a> {
```
alessandrod: nit: I'd call this BufferInput, FileInput contains bytes too

src/linker.rs
```diff
@@ -141,6 +137,58 @@ pub enum OptLevel {
...
+impl<'a> Read for LinkerInput<'a> {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        match self {
+            LinkerInput::File(input) => input.file.read(buf),
+            LinkerInput::Bytes(input) => input.cursor.read(buf),
+        }
+    }
+}
+
+impl<'a> From<(&'a str, &'a [u8])> for LinkerInput<'a> {
```
alessandrod: This doesn't look great. Now that there are two ways to create LinkerInput, I'd
make it new_from_file and new_from_buffer

src/linker.rs
```diff
@@ -645,3 +497,347 @@ fn detect_input_type(data: &[u8]) -> Option<InputType> {
...
+    inner: MemoryBuffer,
+}
+
+impl LinkerOutput {
+    pub fn as_slice(&self) -> &[u8] {
+        self.inner.as_slice()
+    }
+}
+
+impl AsRef<[u8]> for LinkerOutput {
```
alessandrod, line 899: Deref?

src/bin/bpf-linker.rs
banditopazzo: ok got it, it is because of rustc, I will revert it

I removed it because it was unused

src/llvm/types/context.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub(super) context: LLVMContextRef,
+}
+
+impl LLVMContext {
+    pub unsafe fn new() -> Self {
+        let context = LLVMContextCreate();
+        Self { context }
+    }
+
+    pub(in crate::llvm) unsafe fn as_raw(&self) -> LLVMContextRef {
```
vadorovsky: This method doesn't need to be `unsafe` - it's just returning a raw pointer, which isn't an unsafe operation. Look at the `as_ptr` methods in std - https://doc.rust-lang.org/std/?search=as_ptr - they are all safe.

Further usage of this pointer (that involves dereferencing) might be unsafe, but that's the callers responsibility and the compiler is still going to detect that, even when this method isn't unsafe.

Let's also change the name to `as_ptr`.

src/llvm/types/context.rs
```diff
@@ -0,0 +1,72 @@
...
+        self.context
+    }
+
+    pub unsafe fn create_module<'ctx>(&'ctx self, name: &str) -> Option<LLVMModule<'ctx>> {
```
vadorovsky: This also shouldn't be unsafe - this method wraps the llvm-sys functions safely and there is no way a caller can do anything unsound with it.

src/llvm/types/context.rs
```diff
@@ -0,0 +1,72 @@
...
+            return None;
+        }
+
+        Some(LLVMModule {
+            module,
+            _marker: PhantomData,
+        })
+    }
+
+    pub unsafe fn set_diagnostic_handler<T: LLVMDiagnosticHandler>(&self, handler: &mut T) {
```
vadorovsky: Same here

src/llvm/types/module.rs
```diff
@@ -0,0 +1,85 @@
...
+
+use crate::llvm::{types::context::LLVMContext, MemoryBuffer, Message};
+
+pub struct LLVMModule<'ctx> {
+    pub(super) module: LLVMModuleRef,
+    pub(super) _marker: PhantomData<&'ctx LLVMContext>,
+}
+
+impl<'ctx> LLVMModule<'ctx> {
+    pub(in crate::llvm) unsafe fn as_raw(&self) -> LLVMModuleRef {
```
vadorovsky: Same here

src/llvm/types/module.rs
```diff
@@ -0,0 +1,85 @@
...
+        self.module
+    }
+
+    pub unsafe fn get_target(&self) -> *const c_char {
```
vadorovsky: Again, no need for `unsafe`. Also, we could do the conversion to string here already instead of having the caller do it.

src/llvm/types/module.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    pub fn write_bitcode_to_file(&self, output: &CStr) -> Result<(), String> {
+        if unsafe { LLVMWriteBitcodeToFile(self.module, output.as_ptr()) } == 1 {
+            return Err("failed to write bitcode".to_string());
+        }
+
+        Ok(())
+    }
+
+    pub unsafe fn write_bitcode_to_memory(&self) -> MemoryBuffer {
```
vadorovsky: Also no need for unsafe, unless there is any way it can do something unsound depending on the state of `self`?

src/llvm/types/module.rs
```diff
@@ -0,0 +1,85 @@
...
+        let buf = llvm_sys::bit_writer::LLVMWriteBitcodeToMemoryBuffer(self.module);
+
+        MemoryBuffer { memory_buffer: buf }
+    }
+
+    pub unsafe fn write_ir_to_file(&self, output: &CStr) -> Result<(), String> {
+        let (ret, message) =
+            Message::with(|message| LLVMPrintModuleToFile(self.module, output.as_ptr(), message));
```
vadorovsky: To make it more rusty, let's take `AsRef<Path>`, so callers don't have to deal with type conversion.

```suggestion
    pub fn write_ir_to_file<P>(&self, output: P) -> Result<(), String>
    where
        P: AsRef<Path>,
    {
        let output = output
            .as_ref()
            .as_os_str()
            .as_encoded_bytes()
            .as_ptr()
            .cast();
        let (ret, message) =
            unsafe { Message::with(|message| LLVMPrintModuleToFile(self.module, output, message)) };
```

src/llvm/types/module.rs
```diff
@@ -0,0 +1,85 @@
...
+        if ret == 0 {
+            Ok(())
+        } else {
+            Err(message.as_c_str().unwrap().to_str().unwrap().to_string())
+        }
+    }
+
+    pub unsafe fn write_ir_to_memory(&self) -> MemoryBuffer {
```
vadorovsky: I think this also doesn't have to be unsafe

src/llvm/types/module.rs
```diff
@@ -0,0 +1,85 @@
...
+            bytes.len(),
+            buffer_name.as_ptr(),
+        );
+        LLVMDisposeMessage(ptr);
+
+        MemoryBuffer { memory_buffer }
+    }
+
+    /// strips debug information, returns true if DI got stripped
+    pub unsafe fn strip_debug_info(&mut self) -> bool {
```
vadorovsky: Same here

src/llvm/types/target_machine.rs
```diff
@@ -0,0 +1,98 @@
...
+};
+
+use crate::llvm::{types::module::LLVMModule, MemoryBuffer, Message};
+
+pub struct LLVMTargetMachine {
+    pub(super) target_machine: LLVMTargetMachineRef,
+}
+
+impl LLVMTargetMachine {
+    pub unsafe fn new(
```
vadorovsky: Same here

src/llvm/types/target_machine.rs
```diff
@@ -0,0 +1,98 @@
...
+            LLVMCodeModel::LLVMCodeModelDefault,
+        );
+        if tm.is_null() {
+            None
+        } else {
+            Some(LLVMTargetMachine { target_machine: tm })
+        }
+    }
+
+    pub(in crate::llvm) unsafe fn as_raw(&self) -> LLVMTargetMachineRef {
```
vadorovsky: No unsafe and `as_ptr`

src/llvm/types/target_machine.rs
```diff
@@ -0,0 +1,98 @@
...
+    ) -> Result<(), String> {
+        let (ret, message) = Message::with(|message| {
+            LLVMTargetMachineEmitToFile(
+                self.target_machine,
+                module.module,
+                output.as_ptr() as *mut _,
+                output_type,
+                message,
+            )
+        });
```
vadorovsky: Let's take `AsRef<Path>` and make it safe

```suggestion
    pub fn codegen_to_file<P>(
        &self,
        module: &LLVMModule,
        output: P,
        output_type: LLVMCodeGenFileType,
    ) -> Result<(), String>
    where
        P: AsRef<Path>,
    {
        let output = output
            .as_ref()
            .as_os_str()
            .as_encoded_bytes()
            .as_ptr()
            .cast();
        let (ret, message) = unsafe {
            Message::with(|message| {
                LLVMTargetMachineEmitToFile(
                    self.target_machine,
                    module.module,
                    output,
                    output_type,
                    message,
                )
            })
        };
```

src/llvm/types/target_machine.rs
```diff
@@ -0,0 +1,98 @@
...
+        if ret == 0 {
+            Ok(())
+        } else {
+            Err(message.as_c_str().unwrap().to_str().unwrap().to_string())
+        }
+    }
+
+    pub unsafe fn codegen_to_mem(
```
vadorovsky: Unnecessary unsafe?

src/linker.rs
```diff
@@ -141,6 +137,58 @@ pub enum OptLevel {
     SizeMin,
 }
 
+pub struct FileInput<'a> {
+    path: &'a Path,
+    file: File,
```
vadorovsky: I think it would be better to keep only the `path` here and open the file when we are actually processing the input.

Currently, because of storing the `File` instance in `FileInput`, the code in `bpf-linker.rs` looks like:

```rust
    let inputs: Vec<LinkerInput> = inputs
        .iter()
        .map(|p| LinkerInput::try_from(p.as_path()))
        .collect::<Result<_, _>>()?;
```

It has to do an allocation just to handle the `Result`, where error can happen if a file cannot be opened.

If we:

* keep just a `Path` in `FileInput`
* handle opening the file somewhere else
* apply @alessandrod's comment about taking `impl IntoIterator<Item = impl BorrowMut<LinkerInput<'a>>>`

that code could become:

```rust
    let inputs: Vec<LinkerInput> = inputs
        .iter()
        .map(|p| LinkerInput::from(p.as_path()));
```

src/bin/bpf-linker.rs
vadorovsky: Pls add the docstring back as well

banditopazzo: done

src/linker.rs
```diff
@@ -77,6 +69,10 @@ pub enum LinkerError {
     /// The input object file does not have embedded bitcode.
     #[error("no bitcode section found in {0}")]
     MissingBitcodeSection(PathBuf),
+
+    /// LLVM cannot create a module for linking.
+    #[error("failed to create module")]
+    ModuleCreationError,
```
alessandrod: For consistency this should probably be CreateModuleError

src/linker.rs
```diff
@@ -229,354 +288,183 @@ pub struct LinkerOptions {
...
+    /// #     disable_expand_memcpy_in_order: false,
+    /// #     disable_memory_builtins: false,
+    /// #     allow_bpf_trap: false,
+    /// #     btf: false,
+    /// # };
+    /// # let linker = Linker::new(options)?;
+    ///
+    /// // Link into an in-memory buffer.
+    /// let out_buf = linker.link_to_buffer(
+    ///     vec![
```
alessandrod: nit: vec not necessary anymore

src/linker.rs
```diff
@@ -229,354 +288,183 @@ pub struct LinkerOptions {
...
+    /// println!("Linked {} bytes into memory)", bytes.len());
+    ///
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn link_to_buffer<'i>(
+        &self,
+        inputs: impl IntoIterator<Item = LinkerInput<'i>>,
+        output_type: OutputType,
+        export_symbols: &HashSet<Cow<'static, str>>,
```
alessandrod: any reason export_symbols and dump_module are passed here and not as part of
LinkerOptions? 

Making inputs and output_type arguments here make sense if we add support for
emitting multiple formats (eg you might want to emit both .ir and .asm for debugging)

src/llvm/types/context.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    /// Returns an unsafe mutable pointer to the LLVM context.
+    ///
+    /// The caller must ensure that the [LLVMContext] outlives the pointer this
+    /// function returns, or else it will end up dangling.
+    pub(in crate::llvm) const fn as_mut_ptr(&self) -> LLVMContextRef {
+        self.context
+    }
+
+    pub fn create_module<'ctx>(&'ctx self, name: &str) -> Option<LLVMModule<'ctx>> {
```
alessandrod: nit: API wise seems a bit weird that we have LLVMContext::new() but instead of
having LLVMModule::new() we have LLVMContext::create_module

src/llvm/types/module.rs
```diff
@@ -0,0 +1,103 @@
...
+        };
+        if ret == 0 {
+            Ok(())
+        } else {
+            Err(message.as_c_str().unwrap().to_str().unwrap().to_string())
+        }
+    }
+
+    pub fn write_ir_to_memory(&self) -> MemoryBuffer {
+        // NOTE: This function implementation wraps the result into a MemoryBuffer,
```
alessandrod: I'd change this comment to:

```
// Format the module to a string, then copy into a MemoryBuffer. We do the extra copy to keep the
// internal API simpler, as all the other codegen methods output a MemoryBuffer.
```

src/llvm/types/target_machine.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    /// Returns an unsafe mutable pointer to the LLVM target machine.
+    ///
+    /// The caller must ensure that the [LLVMTargetMachine] outlives the pointer this
+    /// function returns, or else it will end up dangling.
+    pub(in crate::llvm) const fn as_mut_ptr(&self) -> LLVMTargetMachineRef {
+        self.target_machine
+    }
+
+    pub fn codegen_to_path(
```
alessandrod: nit: call this emit_to_file like the underlying LLVM API?

src/llvm/types/target_machine.rs
```diff
@@ -0,0 +1,110 @@
...
+            })
+        };
+        if ret == 0 {
+            Ok(())
+        } else {
+            Err(message.as_c_str().unwrap().to_str().unwrap().to_string())
+        }
+    }
+
+    pub fn codegen_to_mem(
```
alessandrod: nit: name emit_to_memory_buffer like the underlying API?

alessandrod (commented):
Sorry for the delay! I just finished my crazy month of traveling, so I'll be more responsive from now.

It's looking pretty good, just a couple of API comments

vadorovsky:
No comments besides what @alessandrod said. The wrappers look good now, thanks for addressing my previous comments. :slightly_smiling_face: 

There is a huge chance that the CI failures will disappear after you rebase (beta switched to newer LLVM, 1.86.0 looks like a flake).

src/linker.rs
banditopazzo: the reason is because now the `Linker` object is reusable, so the symbols to export may change depending on the needs. instead regarding `dump_module`, you don't want to overwrite the pre and post optimization files if you call `link_` functions multiple times.

for example my use case is something like this:
```rust
let linker = Linker::new(LinkerOptions { 
  // ...
})?;

for item in stuff {
  // ...
  let export_symbols = todo!(); // calculate export symbols
  let dump_module_path = todo!(); // optional dump path
  // ...
  let result = linker.link_to_buffer(
        inputs,
        output_type,
        &export_symbols,
        dump_module_path.as_deref(),
  )?;
  // ...
}
```

the symbols I have to export are different for every iteration and I want a different path for each `link_*` call in case I have to look at the ir.

I agree that maybe it should be improved, but right now this PR is too big, I think it's better to split in other tasks.

I have few other stuff I want to add but after this

src/llvm/types/context.rs
banditopazzo: The reason is because the module now has a lifetime tied to the context. the idea for this api comes from `inkwell` https://thedan64.github.io/inkwell/inkwell/context/struct.Context.html#method.create_module

the alternative will be `Module::new(ctx)` but honestly I prefer the first

vadorovsky: I also prefer `create_module` method. This API feels intuitive to me and makes it clear that modules are created only inside contexts, never just on their own out of nowhere.

LLVM C++ API implements [an `LLVMContext` constructor that takes `LLVMContext &C` as an argument](https://llvm.org/doxygen/classllvm_1_1Module.html#a378f93ece2ac999e500f07056cfe6528), but I actually find it confusing and inconsistent, because then many other entities inside `LLVMContext` and `Module` are created by methods like:

* [`LLVMContext::getOrInsertBundleTag`](https://llvm.org/doxygen/classllvm_1_1LLVMContext.html#a624ccf92d0c47b9a2da6534870b007b0)
* [`LLVMContext::getOrInsertSyncScopeID`](https://llvm.org/doxygen/classllvm_1_1LLVMContext.html#adc484415b469ae5d438c3fd803ca3a4f)
* [`Module::getOrInsertFunction`](https://llvm.org/doxygen/classllvm_1_1Module.html#abb107e4edbf05eae92936cba6801c2d9)
* [`Module::getOrInsertGlobal`](https://llvm.org/doxygen/classllvm_1_1Module.html#ae2fcaead977b30daccedc4e397d5e1f0)
* [`Module::getOrInsertNamedMetadata`](https://llvm.org/doxygen/classllvm_1_1Module.html#ab33e5587b25f9c80f6ea5000124d1a2b)

etc.

vadorovsky:
@banditopazzo I pushed a commit with a fix. The problem was that the C strings were not null terminated and apparently LLVM 19 expects the null character. It's a good practice to add it for all FFI calls anyways. Sorry for suggesting a faulty path conversion in my previous comments!

src/linker.rs
alessandrod: > the reason is because now the `Linker` object is reusable, so the symbols to export may change depending on the needs. instead regarding `dump_module`, you don't want to overwrite the pre and post optimization files if you call `link_` functions multiple times.
> 
> for example my use case is something like this:
> 
> ```rust
> let linker = Linker::new(LinkerOptions { 
>   // ...
> })?;
> 
> for item in stuff {
>   // ...
>   let export_symbols = todo!(); // calculate export symbols
>   let dump_module_path = todo!(); // optional dump path
>   // ...
>   let result = linker.link_to_buffer(
>         inputs,
>         output_type,
>         &export_symbols,
>         dump_module_path.as_deref(),
>   )?;
>   // ...
> }
> ```
> 
> the symbols I have to export are different for every iteration and I want a different path for each `link_*` call in case I have to look at the ir.

Yes, but you can also do this by doing 
```
linker.options.dump_module_path = todo!();
let res = linker.link_to_buffer(inputs, output_type, exported_symbols);
```

When designing an API, you have to think what's the most common case, and what's the cognitive overhead of leaking more advanced fringe cases to the common API. You have convinced me that passing `export_symbols` to link is correct, but `dump_module_path` is definitely a fringe case, and shouldn't pollute the API. 

Also in your example you've made it `&HashSet::new()` which is incorrect - if you don't export any symbols DCE will remove all the code, please fix that. 

> I agree that maybe it should be improved, but right now this PR is too big, I think it's better to split in other tasks.
> 
> I have few other stuff I want to add but after this

This is the central point of this PR, and it's a public API, so we have to get it as right as possible and we should aim to break API as little as possible.

src/llvm/types/context.rs
alessandrod: ok, I don't feel strongly about this

alessandrod (changes requested):
See https://github.com/aya-rs/bpf-linker/pull/295#discussion_r2384601821

That's the only thing left then it's ready to go

src/llvm/types/module.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    pub fn write_bitcode_to_memory(&self) -> MemoryBuffer {
+        let buf = unsafe { llvm_sys::bit_writer::LLVMWriteBitcodeToMemoryBuffer(self.module) };
+
+        MemoryBuffer { memory_buffer: buf }
+    }
+
+    pub fn write_ir_to_path(&self, path: impl AsRef<Path>) -> Result<(), String> {
+        let path = CString::new(path.as_ref().as_os_str().as_encoded_bytes()).unwrap();
```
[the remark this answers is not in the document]
vadorovsky: nah

src/linker.rs
```diff
@@ -645,3 +533,354 @@ fn detect_input_type(data: &[u8]) -> Option<InputType> {
...
+    }
+    args.extend(options.llvm_args.iter().map(Into::into));
+    info!("LLVM command line: {:?}", args);
+
+    unsafe {
+        llvm::init(&args, "BPF linker");
+    }
+
+    let context = LLVMContext::new();
+    context.set_diagnostic_handler(&mut diagnostic_handler);
```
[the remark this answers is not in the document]
vadorovsky: You could use `Pin` and `Box` for the diagnostic handler, like @tamird did here:

https://github.com/aya-rs/bpf-linker/pull/307/files#diff-2ae8d7b12653fc1f1413b12c532f72291f5caeea0afa0257d31091b03015eafaR235

Since you're adding a safe wrapper, I think it makes sense to make it fully safe already.

Given that your PR is older, required a lot of work and is pretty close to being accepted by everyone, I think it's better to get in yours first. Pinning is probably the only critical issue that linters found in @tamird's PR.

tamird:
I landed quite a large change in #307 which might be a pain to rebase, sorry about that. On the bright side it removes many unwraps.

src/linker.rs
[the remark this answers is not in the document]
vadorovsky: Uh, ok, that PR actually got merged, sorry :/

tamird:
The merge conflict here is unfortunately very large due to all the code movement included in this PR. IMO it would be better to separate functional changes from code movement both to ameliorate this problem and to ease code review.

src/llvm/mod.rs
```diff
@@ -65,18 +64,8 @@ pub(crate) fn init(args: &[Cow<'_, CStr>], overview: &CStr) {
     unsafe { LLVMParseCommandLineOptions(c_ptrs.len() as i32, c_ptrs.as_ptr(), overview.as_ptr()) };
 }
 
-pub(crate) fn create_module(name: &CStr, context: LLVMContextRef) -> Option<LLVMModuleRef> {
```
tamird, line 68: here is an example of an API I had changed (to accept CStr) to avoid an unwrap, which is lost in this rebase.

banditopazzo: yeah, it was very difficult to rebase, I had to constantly look at your changes. I have to double check everything, this was my first round. 

And I would like to also find a safer solution for the diagnostic handler. Pin locks the memory address but doesn't ensure it remains alive

tamird: Perhaps it would be easier to separate the changes as I suggested earlier: either avoid all the code movement in this PR and do it in a later one or do the code movement first in a separate PR and then rebase this.

banditopazzo: sorry for this time, I went too far with pr. In the beginning I tried to split that in 2 but it was even more because of temporary compatibility, next PRs I promise will be more concise

src/linker.rs
tamird: Looks like this no longer takes the dump path, but if it's per-operation then it makes sense to me that it would go here rather than having to mutate the linker.

alessandrod: Disagree!

If we're serious about making bpf-linker a lib crate (and we should), the API must be nice.

In the vast majority of the cases when you're linking something you aren't debugging LLVM (you are linking), so it doesn't make sense that you have to effectively always pass a `None` to the main API.

In the fringe cases where you're debugging LLVM, we can teach you to mutate something somewhere else.

banditopazzo: I agree with alessandrod on this after he pointed it out. the `None` is basically the default and it's just noise on the API

src/linker.rs
tamird: yeah, slice - or better yet, `IntoIter<Item=Path>`

src/llvm/types/module.rs
tamird: on the other hand we now have 2 copies of this expression containing an unwrap. imo it's better in the caller, especially as this is not public API

src/llvm/types/target_machine.rs
tamird: ditto, I think better in caller

src/llvm/types/module.rs
vadorovsky: Hmm... in the current shape, where all the `write_*` and `emit_*` methods are being called only in [`codegen_to_file`](https://github.com/aya-rs/bpf-linker/blob/eda9d591613984846975b66daaf4af6f6d915cab/src/linker.rs#L733-L754), we could indeed to the conversion there, once for all output types:

```rust
fn codegen_to_file(
    module: &LLVMModule<'_>,
    target_machine: &LLVMTargetMachine,
    output: &Path,
    output_type: OutputType,
) -> Result<(), LinkerError> {
    info!("writing {:?} to {:?}", output_type, output);
    let output = CString::new(output
        .as_os_str()
        .as_encoded_bytes()
        .as_ptr()
        .cast());
    match output_type {
        OutputType::Bitcode => module
            .write_bitcode_to_path(output)
            .map_err(LinkerError::WriteBitcodeError),
        OutputType::LlvmAssembly => module
            .write_ir_to_path(output)
            .map_err(LinkerError::WriteIRError),
        OutputType::Assembly => target_machine
            .emit_to_file(module, output, LLVMCodeGenFileType::LLVMAssemblyFile)
            .map_err(LinkerError::EmitCodeError),
        OutputType::Object => target_machine
            .emit_to_file(module, output, LLVMCodeGenFileType::LLVMObjectFile)
            .map_err(LinkerError::EmitCodeError),
    }
}
```

I think at the time I was making that comment, the calls of these functions were spread in different places and the `CString` construction was done there, overlaping with the actual logic too much.

But honestly, I'm also fine with leaving as it is.

tamird: @banditopazzo would you mind doing this please? this is all private API, so i think it would be better to have one fewer unwrap.

tamird (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/295)* status: all files reviewed, 6 unresolved discussions (waiting on @alessandrod, @banditopazzo, and @vadorovsky)

banditopazzo (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/295)* status: all files reviewed, 6 unresolved discussions (waiting on @alessandrod, @tamird, and @vadorovsky)

___
*[`Cargo.toml` line 37 at r66](https://reviewable.io/reviews/aya-rs/bpf-linker/295#-Oc1p-1z0TvaaZpa_qxu:-Oc2-Ygf2cJJIuYa59oW:b-896fix) ([raw file](https://github.com/aya-rs/bpf-linker/blob/8610dcd283e3546b894bf10aa4b33b7b455b4d55/Cargo.toml#L37)):*

let's not bring in a dependency for this
</blockquote></details>

Done.
___
*[`src/linker.rs` line 690 at r70](https://reviewable.io/reviews/aya-rs/bpf-linker/295#-Oc1p_Ay8IvmYO5Q__nL:-Oc2-sHV9Ggo2tOZux5s:byakpug) ([raw file](https://github.com/aya-rs/bpf-linker/blob/62b4466caf6d4d0f9ed4554a591cc94d8f3bbc54/src/linker.rs#L690)):*

this...does not need to be owned.
</blockquote></details>

I think it's not possible with the previous function signature. The only way I found to avoid the allocation is to change the signature taking `IntoIterator<Item = &'a str>,`
___
*[`src/bin/bpf-linker.rs` line 274 at r70](https://reviewable.io/reviews/aya-rs/bpf-linker/295#-Oc1pnzP7zso4Ni4v1Zt:-Oc2-cg0EZ87rw_ZAjAj:b-896fix) ([raw file](https://github.com/aya-rs/bpf-linker/blob/62b4466caf6d4d0f9ed4554a591cc94d8f3bbc54/src/bin/bpf-linker.rs#L274)):*

I think you just resolved this TODO.
</blockquote></details>

Done.
___
*[`src/llvm/types/module.rs` line 42 at r64](https://reviewable.io/reviews/aya-rs/bpf-linker/295#-Oc1HQSj49cmhuyLsIni:-Oc2-Tt47X4Zl7bwH_ps:b-896fix) ([raw file](https://github.com/aya-rs/bpf-linker/blob/f9a5b4d234e39f86ea72694ee1ec6acf7a8f190e/src/llvm/types/module.rs#L42)):*

https://github.com/lambda-fairy/rust-errno?tab=readme-ov-file#comparison-with-stdioerror

can we just use https://doc.rust-lang.org/std/io/struct.Error.html#method.last_os_error and return `std::io::Error` instead of `String`?
</blockquote></details>

Done.

src/llvm/types/module.rs
banditopazzo: Done.

src/llvm/types/target_machine.rs
banditopazzo: Done.

src/linker.rs
```diff
@@ -232,367 +271,558 @@ pub struct LinkerOptions {
...
+    let mut module = context
+        .create_module(c"linked_module")
+        .ok_or(LinkerError::CreateModuleError)?;
+
+    // buffer used to perform file type detection
+    let mut buf = [0u8; 8];
+    for mut input in inputs {
+        let path = match input {
+            InputReader::File { path, .. } => path.into(),
+            InputReader::Buffer { name, .. } => PathBuf::from(format!("in_memory::{}", name)),
```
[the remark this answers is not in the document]
tamird: This suggestion is plausible.

vadorovsky:
I've restarted the failing job, it it fails everytime with:

```
cargo:warning=error: failed to write /tmp/rustczv1z8A/lib.rmeta: No space left on device (os error 28)
cargo:warning=
cargo:warning=
error: could not compile `rustix` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
cargo:warning=error: failed to build archive at `/home/runner/work/bpf-linker/bpf-linker/aya/target/release/deps/libaya_obj-dee67e733d80af3e.rlib`: No space left on device (os error 28)
cargo:warning=
cargo:warning=
error: could not compile `aya-obj` (lib) due to 1 previous error
rustc-LLVM ERROR: IO failure on output stream: No space left on device
```

And the latest build on `main` also complains about the lack of space:

https://github.com/aya-rs/bpf-linker/actions/runs/18672411715/job/53235931949

I'll try to fix in a separate PR, but I think it's safe to merge this one, given that all the other jobs also run the integration tests.

src/linker.rs
```diff
@@ -232,367 +271,558 @@ pub struct LinkerOptions {
...
+            std::fs::create_dir_all(path)
+                .map_err(|err| LinkerError::IoError(path.to_owned(), err))?;
         }
-        if let Some(path) = &self.options.dump_module {
+        if let Some(path) = dump_module {
             // dump IR before optimization
             let path = path.join("pre-opt.ll");
-            let path = CString::new(path.as_os_str().as_bytes()).unwrap();
-            self.write_ir(&path)?;
+            let path = CString::new(path.as_os_str().as_encoded_bytes()).unwrap();
```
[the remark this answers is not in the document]
vadorovsky: Meh, ignore, let's do it separately

src/linker.rs
```diff
@@ -232,367 +271,558 @@ pub struct LinkerOptions {
...
 
-        Ok(())
+fn codegen_to_file(
+    module: &LLVMModule<'_>,
+    target_machine: &LLVMTargetMachine,
+    output: &Path,
+    output_type: OutputType,
+) -> Result<(), LinkerError> {
+    info!("writing {:?} to {:?}", output_type, output);
+    let output = CString::new(output.as_os_str().as_encoded_bytes()).unwrap();
```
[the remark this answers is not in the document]
vadorovsky: Meh, ignore, let's do it separately

src/linker.rs
```diff
@@ -232,367 +271,558 @@ pub struct LinkerOptions {
...
+    /// #     disable_expand_memcpy_in_order: false,
+    /// #     disable_memory_builtins: false,
+    /// #     allow_bpf_trap: false,
+    /// #     btf: false,
+    /// # };
+    /// # let linker = Linker::new(options);
+    ///
+    /// let export_symbols = ["my_sym_1", "my_sym_2"];
+    ///
+    /// linker.link_to_file(
```
alessandrod, line 330: funniest method name

alessandrod: lmao I'm so dumb I swear apparently I accidentally put the emoji there while reviewing!? SORRY

vadorovsky: Bro, I ended up staring at it for like a minute, trying to understand your comment and spotting shapes/patterns in these letters. I eventually gave up, but later than I shoud've.

PR #343, opened 2026-02-01
build: Ensure that linked libraries are of expected architecture

vadorovsky:
On multiarch environments, simply searching for libraries matching by name might pull in libraries incompatible with the target we are building for.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/343)

PR #353, opened 2026-03-25
build: Support multilib environments

vadorovsky:
Supersedes #343

---

Support systems that provide libraries for multiple address sizes, usually 64-bit and 32-bit, to allow 32-bit applications on 64-bit hosts. Prefer directories with an explicit address size, such as `lib32` or `lib64` for the target, then fall back to `lib`.

Emit warnings that make the directory selection process explicit. If the target-specific directory is missing and `lib` does not contain libraries for the target architecture, the linker will still fail to find them, but the warnings should help debug that case.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/353)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -248,17 +248,32 @@ jobs:
         platform:
           - os: macos-latest
             target: aarch64-apple-darwin
+            # `rustc-llvm-proxy` does not work reliably on macOS. Stable Rust
+            # toolchains for macOS link `libLLVM` statically. Nightly tried to
+            # switch to dynamic linking[0][1], but those attempts were reverted
+            # twice[2][3], which makes the behavior uncertain.
+            #
+            # [0] https://github.com/rust-lang/rust/pull/152768
+            # [1] https://github.com/rust-lang/rust/pull/151063
+            # [2] https://github.com/rust-lang/rust/pull/153079
+            # [3] https://github.com/rust-lang/rust/pull/154485
+            exclude-features: default,rust-llvm-20,rust-llvm-21,rust-llvm-22
           - os: macos-15-intel
             target: x86_64-apple-darwin
+            exclude-features: default,rust-llvm-20,rust-llvm-21,rust-llvm-22
           # We don't use ubuntu-latest because we care about the apt packages available.
           - os: ubuntu-22.04
             target: x86_64-unknown-linux-gnu
           - os: ubuntu-22.04
             target: x86_64-unknown-linux-musl
+            # `rustc-llvm-proxy` does not work on musl targets because Rust
+            # toolchains for musl link libLLVM statically.
+            exclude-features: default,rust-llvm-20,rust-llvm-21,rust-llvm-22
           - os: ubuntu-22.04-arm
             target: aarch64-unknown-linux-gnu
           - os: ubuntu-22.04-arm
             target: aarch64-unknown-linux-musl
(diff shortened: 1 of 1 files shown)


vadorovsky:
Yep, that works:

```diff
diff --git a/llvm/cmake/modules/AddLLVM.cmake b/llvm/cmake/modules/AddLLVM.cmake
index d938214f9d0d..83ef96356dc7 100644
--- a/llvm/cmake/modules/AddLLVM.cmake
+++ b/llvm/cmake/modules/AddLLVM.cmake
@@ -701,7 +701,7 @@ function(llvm_add_library name)

     # Set SOVERSION on shared libraries that lack explicit SONAME
     # specifier, on *nix systems that are not Darwin.
-    if(UNIX AND NOT APPLE AND NOT ARG_SONAME)
+    if(UNIX AND NOT ARG_SONAME)
       set_target_properties(${name}
         PROPERTIES
         # Since 18.1.0, the ABI version is indicated by the major and minor version.
```

```
lrwxr-xr-x    1 vad  staff         31 Mar 27 10:08 libLLVM-22-bpf-linker-0.dylib -> libLLVM.22.1-bpf-linker-0.dylib
-rwxr-xr-x    1 vad  staff   70060160 Mar 27 10:07 libLLVM.22.1-bpf-linker-0.dylib
lrwxr-xr-x    1 vad  staff         31 Mar 27 10:08 libLLVM.dylib -> libLLVM.22.1-bpf-linker-0.dylib
```

Then I'm able to just remove the `libLLVM.dylib` file and tests are passing.

Also, I'm dumb - the reason why I couldn't reproduce the issue locally in https://github.com/aya-rs/bpf-linker/pull/353#issuecomment-4133553034 is that I misspelled the variable for dynamic loading (it's `DYLD_LIBRARY_PATH`, not `DYLIB...`). Now I was able to reproduce the issue and confirm that the fix works.

BTW, which model did you use? In my case, codex wasn't as thorough with finding the exact LLVM commits and so (I tried the default gpt-5.4, and then a self-hosted qwen3-coder). Just wondering if it's my skill issue with asking questions, or I'm using wrong models.

vadorovsky (commented):
> @vadorovsky made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/353)* status: all files reviewed, 2 unresolved discussions (waiting on tamird).

___
*[`-- commits` line 171 at r36](https://reviewable.io/reviews/aya-rs/bpf-linker/353#-Op8ITJT0OpJArrQRHV9:-Op8L3swC-qtS_3qZdn8:bey32n4):*

looks like you forgot to add the footnotes defining 0, 1, 2, 3.

though you have these in the code comment, so you could just mention that comment here if you prefer not to repeat it.
</blockquote></details>

Actually let me add them here as well.
___
*[`.github/workflows/ci.yml` line 260 at r36](https://reviewable.io/reviews/aya-rs/bpf-linker/353#-Op8J1Bp0hckIdE_mkjF:-Op8LCL8DUc7YTqLL8-Y:b-yao9um) ([raw file](https://github.com/aya-rs/bpf-linker/blob/1305a7f3e84dafb107948de9265a40228842ae4f/.github/workflows/ci.yml#L260)):*

i'm surprised we have to write down "default" here. do we?
</blockquote></details>

Yes. Otherwise, `cargo hack --feature-powerset` picks the default feature set, even if it contains features that are excluded. Seems to treat `default` like any other regular feature (which makes sense IMO).

PR #357, opened 2026-04-06
docstring update: mention allows-misaligned-mem-access feature

clairechingching:
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/357)

### DIFF
--- a/src/bin/bpf-linker.rs
+++ b/src/bin/bpf-linker.rs
@@ -88,9 +88,10 @@ struct CommandLine {
     #[clap(long, default_value = "generic")]
     cpu: Cpu,
 
-    /// Enable or disable CPU features. The available features are: alu32, dummy, dwarfris. Use
-    /// +feature to enable a feature, or -feature to disable it.  For example
-    /// --cpu-features=+alu32,-dwarfris
+    /// Enable or disable CPU features. The available features are: alu32, dummy, dwarfris.
+    /// LLVM 22 builds also support allows-misaligned-mem-access. Use +feature to enable a
+    /// feature, or -feature to disable it. For example
+    /// --cpu-features=+allows-misaligned-mem-access,+alu32,-dwarfris
     #[clap(long, value_name = "features", default_value = "")]
     cpu_features: CString,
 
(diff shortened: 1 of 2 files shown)


src/bin/bpf-linker.rs
```diff
@@ -88,9 +88,10 @@ struct CommandLine {
     #[clap(long, default_value = "generic")]
     cpu: Cpu,
 
-    /// Enable or disable CPU features. The available features are: alu32, dummy, dwarfris. Use
-    /// +feature to enable a feature, or -feature to disable it.  For example
-    /// --cpu-features=+alu32,-dwarfris
+    /// Enable or disable CPU features. The available features are: alu32, dummy, dwarfris.
+    /// LLVM 22 builds also support allows-misaligned-mem-access. Use +feature to enable a
+    /// feature, or -feature to disable it. For example
+    /// --cpu-features=+allows-misaligned-mem-access,+alu32,-dwarfris
```
vadorovsky, line 94: Given that the available feature set is fixed, and one of them is available only in a specific LLVM version, I'm wondering if we should rework this CLI option and have an enum representing available features and use it in parsing logic.

With an enum, we could guard different variants with LLVM versions if necessary, e.g.:

```rust
#[derive(Copy, Clone, Debug, Eq, PartialEq, clap::ValueEnum)]
enum CpuFeature {
    Alu32,
    Dummy,
    Dwarfris,
    #[cfg(feature = "llvm-22")]
    AllowsMisalignedMemAccess,
}
```

I will post a full example of the change a bit later today.

vadorovsky: Actually, after thinking about it more, handwriting an enum isn't any better. The main issue - which is us remembering to maintain this list and keep it up to date - still remains.

I was wondering if we could use [`MCSubtargetInfo::getAllProcessorFeatures()`](https://llvm.org/doxygen/classllvm_1_1MCSubtargetInfo.html#a7e3e638e179b49526e659a066fba0fb0) to avoid the manual work, but it's not exposed in LLVM C API. Would be nice to add it, maybe we could avoid handwriting the features while switching to LLVM 23.

For now, I'm going to merge this PR.

vadorovsky (approved):

PR #358, opened 2026-04-08
chore: Prepare for release 0.10.3

vadorovsky:
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/358)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ env:
   # Suffix used to distinguish incompatible LLVM build variants.
   # This lets us roll out changes to the LLVM build in a pull request
   # without conflicting with artifacts produced on the main branch.
-  LLVM_BUILD_REV: 2
+  LLVM_BUILD_REV: 3
 
 permissions: {}
 
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -1,0 +1,81 @@
+# Building bpf-linker from source
+
+## LLVM
+
+bpf-linker is a bitcode linker that uses libLLVM to link bitcode inputs. That
+means the LLVM version used by bpf-linker must match the LLVM version used by
+the Rust toolchain you intend to use.
+
+There are two recommended ways of obtaining an appropriate LLVM.
+
+### Building LLVM from source
+
+LLVM can be built from source using the `xtask build-llvm` subcommand, included
+in the bpf-linker sources.
+
+First, clone the LLVM sources from [our fork][llvm-fork], using the branch
+that matches the Rust toolchain you want to use. For example:
+
+```sh
+git clone -b rustc/22.1-2026-01-27 https://github.com/aya-rs/llvm-project ./llvm-project
+```
+
+If in doubt about which branch to use, check the LLVM version used by your Rust
+compiler:
+
+```sh
+rustc [+toolchain] --version -v | grep LLVM
+```
+
+Once the sources are available, LLVM can be built and installed into the
+directory specified by `--install-prefix`, using `--build-dir` to store the
+build state.
+
+```sh
+cargo xtask llvm build \
+    --src-dir ./llvm-project \
+    --build-dir ./llvm-build \
+    --install-prefix ./llvm-install
+```
+
+After that, bpf-linker can be built with the `LLVM_PREFIX` environment
+variable pointing to that directory:
+
+```sh
+export LLVM_PREFIX=./llvm-install
+```
+
+Alternatively, the `bin` directory from the install prefix can be added to
+`PATH`:
+
+```sh
+export PATH="$(pwd)/llvm-install/bin:$PATH"
+```
+
+[llvm-fork]: https://github.com/aya-rs/llvm-project
+
+### System packages
+
+On Debian-based distributions, you can install the `llvm-<version>-dev` and
+`libclang-<version>-dev` packages from the official LLVM repository at
+https://apt.llvm.org.
+
+Different operating systems and Linux distributions might provide their own
+LLVM packages. If you're interested in packaging bpf-linker, you may also need
+to ensure that the correct LLVM version is packaged for that environment.
+
+## Building bpf-linker with Cargo
+
+bpf-linker uses Cargo features to select the LLVM version, via `llvm-*`
+features such as `llvm-22`. By default, LLVM and its dependencies are linked
+dynamically. Static linking can be enabled with the `llvm-link-static` feature.
+
+Examples:
+
+```
+# Dynamic linking, Rust nightly
+cargo +nightly --no-default-features --features llvm-22
+
+# Static linking
+cargo +nightly --no-default-features --features llvm-22,llvm-link-static
+```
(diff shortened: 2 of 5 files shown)


vadorovsky:
Basically the plan is that starting from this release, I want to announce binstalls as the official way(TM)(R) of installing bpf-linker, stop mentioning rustc-llvm-proxy as an option (even though it's still there) and see how the adoption goes. Based on that, we can think when do we want to kill the rustc-llvm-proxy.

tamird (changes requested):
> @tamird partially reviewed 1 file and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/358)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on alessandrod and vadorovsky).

___
*[`README.md` line 21 at r1](https://reviewable.io/reviews/aya-rs/bpf-linker/358#-Opi4mVM2EJvsO5x4Iwj:-Opi4mVM2EJvsO5x4Iwk:bf6rsp9) ([raw file](https://github.com/aya-rs/bpf-linker/blob/fc182e573b5890245b9dfc112da0a4bbd9c31271/README.md#L21)):*
> ```Markdown
> ## Installation
> 
> ### cargo-binstall
> ```

The released linker requires the user to be on the same llvm version linked rustc, right? Iirc there's no easy way to make that actually happen with nightly.

tamird (commented):
> @tamird reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/358)* status: all files reviewed, 1 unresolved discussion (waiting on alessandrod and vadorovsky).

vadorovsky (commented):
> @vadorovsky made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/358)* status: all files reviewed, 1 unresolved discussion (waiting on alessandrod and tamird).

___
*[`README.md` line 21 at r1](https://reviewable.io/reviews/aya-rs/bpf-linker/358#-Opi4mVM2EJvsO5x4Iwj:-OpiAphdCE-jTOUrEcT1:b8p65pj) ([raw file](https://github.com/aya-rs/bpf-linker/blob/fc182e573b5890245b9dfc112da0a4bbd9c31271/README.md#L21)):*

The released linker requires the user to be on the same llvm version linked rustc, right? Iirc there's no easy way to make that actually happen with nightly.
</blockquote></details>

Yeah, I mean, the only way for now is to move fast and chase the latest nightly. Which is what we've been already doing even with rustc-llvm-proxy being the default. :slightly_smiling_face: And I think it's sort of fine as long as for rustup we're stuck in Tier 3, and support of stable Rust toolchains is done only by Linux distributions.

One idea I want to explore for the future, to be ready to support different LLVM versions with prebuilt binaries, is:

* Building different bpf-linker binaries for different LLVM versions, e.g. `bpf-linker-22`, `bpf-linker-21`.
* Having a "proxy" binary `bpf-linker` that parses the version of bitcode and then calls the versioned bin.

But I'll need some time to get it working.

tamird (changes requested):
> @tamird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/358)* status: all files reviewed, 1 unresolved discussion (waiting on alessandrod and vadorovsky).

___
*[`README.md` line 21 at r1](https://reviewable.io/reviews/aya-rs/bpf-linker/358#-Opi4mVM2EJvsO5x4Iwj:-Opik5vCBGdk99YWVQsn:bd2tyb8) ([raw file](https://github.com/aya-rs/bpf-linker/blob/fc182e573b5890245b9dfc112da0a4bbd9c31271/README.md#L21)):*
<details><summary><i>Previously, vadorovsky (Michal R) wrote…</i></summary><blockquote>

Yeah, I mean, the only way for now is to move fast and chase the latest nightly. Which is what we've been already doing even with rustc-llvm-proxy being the default. :slightly_smiling_face: And I think it's sort of fine as long as for rustup we're stuck in Tier 3, and support of stable Rust toolchains is done only by Linux distributions.

One idea I want to explore for the future, to be ready to support different LLVM versions with prebuilt binaries, is:

* Building different bpf-linker binaries for different LLVM versions, e.g. `bpf-linker-22`, `bpf-linker-21`.
* Having a "proxy" binary `bpf-linker` that parses the version of bitcode and then calls the versioned bin.

But I'll need some time to get it working.
</blockquote></details>

Yeah, chasing latest nightly is reasonable, but it isn't what we do now! Because we don't continuously rebuild our releases every time nightly bumps. So if we go on vacation for a while and nightly bumps LLVM then our linker stops working for users.

vadorovsky (commented):
> @vadorovsky made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/358)* status: all files reviewed, 1 unresolved discussion (waiting on alessandrod and tamird).

___
*[`README.md` line 21 at r1](https://reviewable.io/reviews/aya-rs/bpf-linker/358#-Opi4mVM2EJvsO5x4Iwj:-Opkmrw8Am541r70Ka7F:b-tmjvqg) ([raw file](https://github.com/aya-rs/bpf-linker/blob/fc182e573b5890245b9dfc112da0a4bbd9c31271/README.md#L21)):*
> but it isn't what we do now! Because we don't continuously rebuild our releases every time nightly bumps.

Every time nightly bumps LLVM to a new major version, I'm doing the whole ritual of:

* Sending a PR to llvm-sys-rs to support that new LLVM.
* Bumping llvm-sys version.
* Releasing bpf-linker.

You're right that we usually didn't do it for **exactly** every bump of LLVM done by Rust nightly to a new **patch** version. I was always aware of that, the reason I didn't consider it a big issue is that usually patch bumps of LLVM:

* Don't contain any BPF-related changes. There was only one exception - 21.1.8 - which had BPF fixes that I wrote.
* Don't break the bitcode/IR compatibility.

But I'm happy to change that habit just to be safe - we could could pull from https://github.com/rust-lang/llvm-project and cut a patch release every time a new LLVM patch release is pulled by the rust fork. I think that would address your concern.

I guess your point is that rustc-llvm-proxy could still work even if we forget about a major upgrade. But it would work only for a certain group of people (Linux x86_64). Lot of users, including you and Alessandro use macOS. Plenty of Linux users use Alpine or something musl-based. There are folks using compiling for other architectures on Linux. Chasing LLVM updates is the only way to satisfy all of them, that was always the case, and probably always will be.

> So if we go on vacation for a while and nightly bumps LLVM then our linker stops working for users.

That was always the problem, and my weekends/nights are always ruined when a major LLVM version is released. Not sure if we can ever get rid of this problem.

On the other hand, patch bumps are very unlikely to break users.

We could think of some automation to make the process easier. But I don't think we should block this PR and the release because of this problem, which was always there, and I was able to handle it.

tamird (approved):
> @tamird reviewed 3 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/358)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on alessandrod).

vadorovsky:
I need to rebuild LLVM, but I can't from the fork. Superseded by #359

PR #361, opened 2026-04-26
di: Use safe wrappers consistently

vadorovsky:
Avoid usage of `LLVM*Ref` pointer types in favor of safe wrappers. See individual commits.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/361)

### DIFF
--- a/src/llvm/di.rs
+++ b/src/llvm/di.rs
@@ -23,7 +23,7 @@ use crate::llvm::{LLVMContext, LLVMModule, iter::*, types::di::DISubprogram};
 const MAX_KSYM_NAME_LEN: usize = 128;
 
 pub(crate) struct DISanitizer<'ctx> {
-    context: LLVMContextRef,
+    context: &'ctx LLVMContext,
     module: LLVMModuleRef,
     builder: LLVMDIBuilderRef,
     visited_nodes: HashSet<u64>,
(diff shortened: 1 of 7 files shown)


tamird (approved):
> @tamird reviewed 7 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/361)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on alessandrod).

___
*[`src/llvm/types/ir.rs` line 224 at r1](https://reviewable.io/reviews/aya-rs/bpf-linker/361#-OrAg8y44L79cXKuHdFf:-OrAg8y44L79cXKuHdFg:bvvmeq1) ([raw file](https://github.com/aya-rs/bpf-linker/blob/6505cd66f6adeaa845166c28cf6da71f7c5a243d/src/llvm/types/ir.rs#L224)):*
> ```XML
> 
>     /// Constructs an empty metadata node.
>     pub(crate) fn empty(context: &LLVMContext) -> Self {
> ```

a bit funny to pass an immutable ref and call as_mut_ptr on it

i see we do that everywhere though

vadorovsky (commented):
> @vadorovsky made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/361)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on alessandrod).

___
*[`src/llvm/types/ir.rs` line 224 at r1](https://reviewable.io/reviews/aya-rs/bpf-linker/361#-OrAg8y44L79cXKuHdFf:-OrC9ZQH8CSqCWz13176:b707957) ([raw file](https://github.com/aya-rs/bpf-linker/blob/6505cd66f6adeaa845166c28cf6da71f7c5a243d/src/llvm/types/ir.rs#L224)):*

a bit funny to pass an immutable ref and call as_mut_ptr on it

i see we do that everywhere though
</blockquote></details>

You're right, it's a smell - especially since in `std`/`core` all `as_mut_ptr` methods borrow `self` mutably. I will think of a solution and fix that as a follow up.

PR #363, opened 2026-04-28
[WIP] feat: CO-RE/BTF relocations

vadorovsky:
BTF, the BPF Type Format, encodes type information for both the running Linux kernel and compiled eBPF programs. An eBPF object can carry relocation records that describe field and aggregate accesses in terms of BTF types instead of fixed offsets; at load time, the loader compares the program's BTF with the kernel's BTF and rewrites those accesses to the correct offsets for the target kernel. This mechanism is often referred to as "CO-RE relocations" or "BTF relocations".

LLVM allows to emit such relocations with the following intrinsics:

* `@llvm.preserve.array.access.index`
* `@llvm.preserve.struct.access.index`
* `@llvm.preserve.union.access.index`

When these intrinsics are preserved through codegen, LLVM emits the corresponding relocation records into the `BTF.ext` ELF section.

clang exposes this functionality through `__attribute__((preserve_access_index))` that lowers to one of the `@llvm.preserve.*.access.index` intrinsics, depending on the annotated type.

We proposed a similar mechanism for Rust[0], but that will likely take time to land. In the meantime, add support to bpf-linker using the `Relocatable` marker type: when a composite type contains a field of that marker type, treat accesses to that composite type as eligible for CO-RE relocation emission. bpf-linker then constructs the corresponding `@llvm.preserve.*.access.index` intrinsic calls using the existing debug information and lets LLVM emit the final `BTF.ext` relocation records.

[0] https://internals.rust-lang.org/t/pre-rfc-btf-relocations/24161/20

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/363)

### DIFF
--- a/src/linker.rs
+++ b/src/linker.rs
@@ -55,6 +55,11 @@ pub enum LinkerError {
     #[error("LLVMRunPasses failed: {0}")]
     OptimizeError(String),
 
+    /// Rewriting CO-RE relocations failed.
+    #[cfg(feature = "llvm-22")]
+    #[error("failed to rewrite CO-RE relocations: {0}")]
+    CoreRelocError(llvm::CoreRelocError),
+
     /// Generating the BPF code failed.
     #[error("LLVMTargetMachineEmitToFile failed: {0}")]
     EmitCodeError(String),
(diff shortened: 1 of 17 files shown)


vadorovsky:
I'm closing this. I want to ship the solution in Rustc (https://github.com/rust-lang/rust/pull/161107), so far the Rust PR received feedback, and this one didn't. And I don't see any point in bloating bpf-linker.

PR #367, opened 2026-05-02
ir: Add `value_to_message` private helper

vadorovsky:
The previous `value_to_string` closure was available only to the `Debug` implementation for `Value` and performed an unnecessary allocation by calling `to_string()`.

Replace it with a new helper function, `value_to_message`, that is available to the entire `types` module and can therefore be used by other LLVM wrappers as well. Return `Message`, so callers can retrieve a `Cow` by calling `as_string_lossy()`. That avoids allocations when the `Message` holds valid UTF-8.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/367)

### DIFF
--- a/src/llvm/mod.rs
+++ b/src/llvm/mod.rs
@@ -6,6 +6,7 @@ use std::{
     borrow::Cow,
     collections::HashSet,
     ffi::{CStr, CString},
+    fmt,
     os::raw::c_char,
     ptr, slice, str,
 };
@@ -366,6 +367,12 @@ impl Message {
             .map(CStr::to_bytes)
             .map(String::from_utf8_lossy)
             .unwrap_or("<null>".into())
+    }
+}
+
+impl fmt::Debug for Message {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(&self.as_string_lossy())
     }
 }
 
--- a/src/llvm/types/ir.rs
+++ b/src/llvm/types/ir.rs
@@ -32,6 +32,12 @@ pub(crate) fn replace_name(
     unsafe { LLVMReplaceMDNodeOperandWith(value_ref, name_operand_index, name) };
 }
 
+fn value_to_message(value_ref: LLVMValueRef) -> Message {
+    Message {
+        ptr: unsafe { LLVMPrintValueToString(value_ref) },
+    }
+}
+
 #[derive(Clone)]
 pub(crate) enum Value<'ctx> {
     MDNode(MDNode<'ctx>),
@@ -41,25 +47,18 @@ pub(crate) enum Value<'ctx> {
 
 impl std::fmt::Debug for Value<'_> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        let value_to_string = |value| {
-            Message {
-                ptr: unsafe { LLVMPrintValueToString(value) },
-            }
-            .as_string_lossy()
-            .to_string()
-        };
         match self {
             Self::MDNode(node) => f
                 .debug_struct("MDNode")
-                .field("value", &value_to_string(node.value_ref))
+                .field("value", &value_to_message(node.value_ref))
                 .finish(),
             Self::Function(fun) => f
                 .debug_struct("Function")
-                .field("value", &value_to_string(fun.value_ref))
+                .field("value", &value_to_message(fun.value_ref))
                 .finish(),
             Self::Other(value) => f
                 .debug_struct("Other")
-                .field("value", &value_to_string(*value))
+                .field("value", &value_to_message(*value))
                 .finish(),
         }
     }


tamird (approved):
> @tamird reviewed 1 file and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/367)* status: all files reviewed, 2 unresolved discussions (waiting on vadorovsky).

___
*[`src/llvm/types/ir.rs` line 35 at r1](https://reviewable.io/reviews/aya-rs/bpf-linker/367#-Ori7FuN-mhg6iGKSR3O:-Ori7FuN-mhg6iGKSR3P:b-5fktz7) ([raw file](https://github.com/aya-rs/bpf-linker/blob/63b200319299eae10f07441b494a3a512d3c8068/src/llvm/types/ir.rs#L35)):*
> ```XML
> }
> 
> pub(super) fn value_to_message(value_ref: LLVMValueRef) -> Message {
> ```

super because you plan to use elsewhere?
___
*[`src/llvm/types/ir.rs` line 53 at r1](https://reviewable.io/reviews/aya-rs/bpf-linker/367#-Ori7PSmBvZILJhgSsGV:-Ori7PSn9j5Lw421z5i8:b-9y6js2) ([raw file](https://github.com/aya-rs/bpf-linker/blob/63b200319299eae10f07441b494a3a512d3c8068/src/llvm/types/ir.rs#L53)):*
> ```XML
>             Self::MDNode(node) => f
>                 .debug_struct("MDNode")
>                 .field("value", &value_to_message(node.value_ref).as_string_lossy())
> ```

Presumably this derefs to CStr? I wonder if you could .display() instead

vadorovsky (commented):
> @vadorovsky made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/367)* status: all files reviewed, 2 unresolved discussions (waiting on tamird).

___
*[`src/llvm/types/ir.rs` line 35 at r1](https://reviewable.io/reviews/aya-rs/bpf-linker/367#-Ori7FuN-mhg6iGKSR3O:-OrlqBLB2qv7Ow16PYyY:bhni4s) ([raw file](https://github.com/aya-rs/bpf-linker/blob/63b200319299eae10f07441b494a3a512d3c8068/src/llvm/types/ir.rs#L35)):*

super because you plan to use elsewhere?
</blockquote></details>

Yes, it's quite convenient for providing `Debug` implementations for other wrappers, e.g.

https://github.com/aya-rs/bpf-linker/pull/363/changes/dd12121348b3f70d13e016f1259f1a601057bca5#diff-82d6674953aa246d643cd2d918573549505826b8572984758b81f68045619a18R299-R306

That said, I can make it private for now and change the publicity later, when actually adding more wrappers.
___
*[`src/llvm/types/ir.rs` line 53 at r1](https://reviewable.io/reviews/aya-rs/bpf-linker/367#-Ori7PSmBvZILJhgSsGV:-Orlv1Mz68Pw-DwD3jud:b-9bgvfe) ([raw file](https://github.com/aya-rs/bpf-linker/blob/63b200319299eae10f07441b494a3a512d3c8068/src/llvm/types/ir.rs#L53)):*

Presumably this derefs to CStr? I wonder if you could .display() instead
</blockquote></details>

To be precise - this derefs to `Cow<'a, CStr>`. Unfortunately, `CStr::display()` is not stabilized yet: https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.display

But you made me realize that I can still make this prettier by implementing `Debug` for `Message`,  so I don't need these `.as_string_lossy()` calls. See the last push.

tamird (approved):
> @tamird reviewed 2 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/367)* status: all files reviewed, 1 unresolved discussion (waiting on vadorovsky).

___
*[`src/llvm/mod.rs` line 373 at r2](https://reviewable.io/reviews/aya-rs/bpf-linker/367#-OrndUiw6T4GWHeQ_dvK:-OrndUiw6T4GWHeQ_dvL:b-9oli3n) ([raw file](https://github.com/aya-rs/bpf-linker/blob/6053ed79b69457943a064bb712f9001c1746a7b6/src/llvm/mod.rs#L373)):*
> ```XML
> }
> 
> impl fmt::Debug for Message {
> ```

Should this be Display? CStr does impl Debug

https://doc.rust-lang.org/stable/std/ffi/struct.CStr.html#impl-Debug-for-CStr

if you do make it display please add a TODO with CStr Display tracking issue

vadorovsky:
After giving it more thought, I don't think doing it this way makes sense.

`Message` is designed for error handling, where error messages are optional, and a `NULL` `c_char` is a legitimate case, where it means that no error was caught.

In case of `Value`, a `NULL` pointer is a bug and we likely want to have some error. Before jumping into enhancing the `Debug` implementations of LLVM types, it's probably better to make them use `NonNull`, to make sure we don't have to handle `NULL` pointers in random places, but rather error while instantiating the wrappers. I'll make a separate PR for that.

PR #402, opened 2026-08-13
fix(bazel): Report correct version

### LINKED ISSUE #401: Latest release has --version report an invalid version
The latest release of bpf-linker has its `--version` broken as it reports `0.0.0` instead of `0.11.0`

Steps to reproduce
```console
$ docker run --rm -it rust:bullseye
[...]
root@160a305adce8:/# curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | bash
[...]
root@160a305adce8:/# cargo binstall bpf-linker
[...]
root@160a305adce8:/# bpf-linker --version
bpf-linker 0.0.0
```

Its the same thing if I install via bininstall or if I download the github release manually

vadorovsky:
Bazel-built binaries were reporting `--version` as 0.0.0 instead of 0.11.0 because `CARGO_PKG_VERSION` was not set. Add version.bzl module and make sure it's kept in sync with Cargo.toml. Pass the version from version.bzl in `rustc_env`.

Fixes: #401

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/bpf-linker/402)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -174,6 +174,25 @@ jobs:
             oras tag "$sha" "$version"
           done
 
+  check-version:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Generate version.bzl
+        run: ./bazel/version/gen_version_bzl.sh > version.bzl.new
+
+      - name: Check version.bzl is in sync
+        run: |
+          if ! diff -u bazel/version/version.bzl version.bzl.new; then
+            echo "ERROR: version.bzl is out of sync with Cargo.toml"
+            echo "Run: bash bazel/version/gen_version_bzl.sh > bazel/version/version.bzl"
+            exit 1
+          fi
+
   lint-stable:
     runs-on: ubuntu-latest
 
@@ -543,6 +562,7 @@ jobs:
       - bazel
       - build
       - release
+      - check-version
     runs-on: ubuntu-latest
     steps:
       - name: Verify all required jobs succeeded
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -11,6 +11,7 @@ load(
     "release_windows_filegroup",
 )
 load("//bazel:llvm_archive.bzl", "llvm_archive", "llvm_archive_for_platform")
+load("//bazel/version:version.bzl", "BPF_LINKER_VERSION")
 
 # If Cargo.toml remaps any crate names, this tells Bazel how to pass the right
 # `--extern` flags.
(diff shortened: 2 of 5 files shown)


vadorovsky:
@dzbarsky I'm still a Bazel n00b. LLMs say this is the only way, but I'm wondering if we could somehow teach bazel to always honor versions from Cargo.toml without such hacks.

dzbarsky:
> @dzbarsky I'm still a Bazel n00b. LLMs say this is the only way, but I'm wondering if we could somehow teach bazel to always honor versions from Cargo.toml without such hacks.

I don't really like this approach - if you're going to parse the Cargo.toml you can at least do it via `repository_rule` so you don't have to check in the contents. Pro tip - invoke `toml2json` in the repository rule, then parse the json, then write the output to a `.bzl`. This is what rules_rs does already, see https://github.com/hermeticbuild/rules_rs/blob/2bd8e3564b3d55ca5d8c15d3ffa2f8ee8d680819/rs/private/git_cargo_workspace_repository.bzl#L18-L20. This would be my preferred approach!

So something like this:
```
load("@rules_rs//rs/private:toml2json.bzl", "run_toml2json")

def _cargo_toml_version_impl(rctx):
    version = run_toml2json("@//:Cargo.toml")["version"]
    rctx.file("version.bzl", 'VERSION = "%s"' % version)
    return rctx.repo_metadata(reproducible = True)

cargo_toml_version = repository_rule(impl = _cargo_toml_version_impl)

.....MODULE.bazel
use_repo_rule("//path/to/rule.bzl", "cargo_toml_version")
cargo_toml_version(name = "cargo_toml_version") 

....BUILD.bazel
load("@cargo_toml_version//:version.bzl", "VERSION")

# use VERSION in rustc_env
```


In case you're curious, some more options:
- do something like so: https://github.com/bazelbuild/rules_rust/blob/bb4e9e8cf14511e7b9cf7bda3d1c1f277d076e24/cargo/private/cargo_toml_variable_extractor/testdata/BUILD.bazel#L80-L90 . The downside is that it requires compiling the `cargo_toml_var_extractor` binary so it's a little annoying, but not the end of the world if you do want this embedded.
- use [workspace_status_command/stamp](https://bazel.build/docs/user-manual#workspace-status) to embed the version (though you'd still need to put `VERSION=$(grep '^version' "$CARGO_TOML" | sed 's/version = "\([^"]*\)"/\1/')` into the workspace_status_command script to wire it up)

vadorovsky:
@dzbarsky I'm going to be off for a month now, sorry that I didn't get to addressing your comment. Feel free to propose a fix, and close this PR, if you wish.

